PR #27, opened 2013-04-16
Less memsetting in sprite rendering

Nebuleon:
I attach one commit to initialise less memory in the FirstSprite+Y case of sprite rendering. This may help the memory performance of some games that use this rendering case.


BearOso:
Since this is obsolete, I'm closing it.

PR #36, opened 2013-07-19
Some fix and clean-up stuff

linux0r:
Before improving snes9x (in term of features and GUI), my goal is to fix every warning so that we have a clean code.

When I started making changes, I found make targets to be incomplete some I added 'distclean' and updated 'clean' one.
I updated the configure script with a recent version of autoconf and started toying with configure options.
I found shared-memory was used even if dependencies were not met so I fixed that.
When I tried cross-compiling snes9x (to MIPS which is big-endian) to test my endiannes-related changes, I tweaked the configure script to disable some tests (because you can't run the binaries if cross-compiling) and to detect correctly X11.
I also added a 'Cross-compiling' section in readme_unix.html.
Finally, I fixed some warnings like '-Wdelete-non-virtual-dtor' and '-Wstrict-aliasing'.
I know the last changes introduce more code and seems unneeded but cleaner code is always better than some compact (and obfuscated) code.
Anyway, I tested my code (x86_64 and MIPS under qemu) and it work as before.

If these changes don't interest you, let me know.
Thanks

### DIFF
--- a/macosx/mac-global_prefix.h
+++ b/macosx/mac-global_prefix.h
@@ -189,8 +189,6 @@
  ***********************************************************************************/
 
 
-#undef	WRITE_DWORD(s, d)
-
 #define ZLIB
 #define UNZIP_SUPPORT
 #define	JMA_SUPPORT
--- a/port.h
+++ b/port.h
@@ -442,11 +442,23 @@ static inline uint32 READ_DWORD(uint8 *in)
     return (out.u32);
 }
 
-#ifdef FAST_LSB_WORD_ACCESS
-#define WRITE_DWORD(s, d)	*(uint32 *) (s) = (d)
+static inline void WRITE_DWORD(uint8 *out, uint32 in)
+{
+    union DWORD tmp;
+
+    tmp.u32 = in;
+#if defined(LSB_FIRST)
+    out[0] = tmp.u8[0];
+    out[1] = tmp.u8[1];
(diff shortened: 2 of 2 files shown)


vgturtle127:
What is the status of this PR?

s0m30n3:
I retested the master as of today.

~/Projects/snes9x-master/unix$ ./configure
<snip>
checking for X... libraries , headers 
<snip>
build information:
cc...............,,,. gcc
c++.................. g++
options.............. -O3 -fomit-frame-pointer -fno-exceptions -fno-rtti -pedantic -Wall -W -Wno-unused-parameter 
defines.............. -DJOYSTICK_SUPPORT -DZLIB -DUNZIP_SUPPORT -DJMA_SUPPORT -DHAVE_LIBPNG -DHAVE_MKSTEMP -DHAVE_STRINGS_H -DHAVE_SYS_IOCTL_H -DHAVE_STDINT_H -DRIGHTSHIFT_IS_SAR -DUSE_THREADS
libs................. -lz -lpng -lSM -lICE -lX11 -lXext -lpthread
<snip>
configure: creating ./config.status
config.status: creating Makefile

Configure output is now clean and mean.

No XShm issue, compiles fine (though still has many warnings): you can safely close it.

This PR is old and now I now commits should be as minimal as possible so I'll do better next time.
Anyway, thanks for the reminder !

PR #59, opened 2014-06-27
Use of rewind feature (stateman) with GTK.

juhalaukkanen:
Works fine, I have tested with several games.

However in config dialog there seems to be some strange GTK warnings. Probably NUM_EMU_LINKS define is not 100% proper.


BearOso:
This is a good start. The GTK warnings are simple to fix. This can't be enabled universally, though; it's too CPU intensive.

PR #62, opened 2015-01-07
gtk: fix desktop entry mimetype

sebn:
Current application/x-snes-rom is not an official IANA mimetype:
https://www.iana.org/assignments/media-types/application/x-snes-rom

The correct mimetype is application/vnd.nintendo.snes.rom:
https://www.iana.org/assignments/media-types/application/vnd.nintendo.snes.rom

The old mimetype was kept since someone may be using it.


sebn:
Thank you :smile:

PR #64, opened 2015-01-28
gtk: add xBRZ (2x, 3x, 4x)

nmagre:
add xBRZ to GTK.
Multithread works ("Use X threads for filtering and scaling" option).
'copyImage16To32' and 'stretchImage32To16' from 'filter_xbrz.cpp' should be deported somewhere else to be used in both win32 renderer and gtk filter.

screenshot-snes9x preferences


BearOso:
I guess this is fine for now. Though I'm not content with converting to 32-bit and back again.

PR #74, opened 2015-04-30
Settings::QuickSaveExtension for quick load/save savestates

axltxl:
**Summary of changes**
A new option has been added to `snes9x.conf`: `Settings::QuickSaveExtension`. 
This will allow the player/user to use different file extension patterns when loading/saving quick savestates. There are three posible values: 0, 1 and 2, which are correspondent to three different file extension patterns:
- Choosing 0 (`SNES_SSEXT_CLASSIC`, which is actually the default) will make snes9x load/save quick savestates using the usual pattern (`.000`, `.001`, `.002`, ...)
- Choosing 1 (`SNES_SSEXT_FRZ`) will make snes9x load/save quick savestates using the `.frz` pattern (`.00.frz`, `.01.frz`, `.02.frz`, ...). This pattern is the one used by the Mac port and snes9x EX+ (Android) as well.
- Choosing 2 (`SNES_SSEXT_ZSNES`) will make snes9x load/save quick savestates using the ZSNES pattern (.zst, .zs1, .zs2, ...)

**Motivation**: Portability
This feature was inspired by the need I had to seamlessly load and save quick savefiles (which are synced via BTSync) accross all emulators I use (snes9x-gtk on my laptop, snes9x EX+ in my smartphone and the mac port in my macbook). Since they manage different file extensions when it comes to quick savestates, I figured out a way to normalize that so I wouldn't have to deal with savefiles whenever I wanted to continue playing a game I started on another snes9x-based emulator from a certain quick savefile. By implementing this set of changes, I managed a way to easily load/save quick savestates without ever touching/renaming a single one of those, therefore, a seamless cross-emulator gaming :relieved: .

**GTK port changes**
Along with this, there is a series of changes included in the GTK port:
- `Settings::QuickSaveExtension` is mapped to `quicksave_ext` in `snes9x.xml`
- *.frz files are included in the load/save state dialog
- Added ability to set `Settings::QuickSaveExtension` in the 'Preferences' dialog, like so:

I have tested this feature in my Linux box, using both core and gtk port (using either gtk2 or gtk3). If you have any doubts regarding this or any kind of observations, please let me know :relaxed: .
I'm open to amend whatever is necessary.

Cheers! :beer:


bearoso:
This isn't a common use-case and is outdated. Closing.

PR #89, opened 2015-09-09
GTK: Provide UI for 'relative-style' save states.

ichigo-0:
In other words, rather than having one load shortcut and one save
shortcut per slot, this allows you to use arbitrarily many slots with
just four shortcuts: save to the current slot, load from the current
slot, set current slot to the next slot, and set current slot to the
previous slot. The 'direct-style' mechanism continues to be available.

A somewhat separate feature is the that when saving state, instead of
immediately overwriting the old state, the old state will be temporarily
set aside. The most recent such old state is then available for loading.
This can be helpful if the user accidentally saves over the wrong slot,
or simply saves at a bad time.

Also, the number of slots is increased from 9 to 10.

(I would have offered these as three separate patches, but they all
require modifying the Glade UI file, and it is difficult split and merge
changes to that.)

### DIFF
--- a/gtk/src/gtk_control.cpp
+++ b/gtk/src/gtk_control.cpp
@@ -65,6 +65,8 @@ const BindingLink b_links[] =
         { "b_save_6",              "QuickSave006"      },
         { "b_save_7",              "QuickSave007"      },
         { "b_save_8",              "QuickSave008"      },
+        { "b_save_9",              "QuickSave009"      },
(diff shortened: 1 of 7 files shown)


bearoso:
This is now implemented. You have 999 total slots for each game. There's keys to save and load from the current slot, increment and decrements the slot number, and also increment and save at the same time (never save over the old one) and decrement and load at the same time (cycle through state history quickly).

PR #97, opened 2015-11-28
Remove calls from deprecated GTK_STOCK constance.

maxolasersquad:
GTK_STOCK has been deprecated since GTK 3.10. This commit removes those calls in favor of the recommendations at https://developer.gnome.org/gtk3/stable/gtk3-Stock-Items.html


maxolasersquad:
This commit seems like a fairly trivial thing. Are there any concerns about it?

bearoso:
I don't agree with the philosophy regarding the change: it's only concerned with GTK3 going forward, which is a dead end. Plus, this gets rid of the nice button icons that appear on GTK2 (which is the priority version), and doesn't automatically translate.

PR #98, opened 2015-11-28
Fixed various compiler warnings.

maxolasersquad:
This removes a lot of the compiler warnings thrown when compiling against GCC. I am not a C++ programmer so please feel free to comment if any of these changes are not optimal or have better solutions.


bearoso:
There are a couple problems with your patch, and the rest are reversions that handle another, different compiler warning (ignore return value). The proper solution if you're concerned is likely to just disable the appropriate warning, but obviously we keep all warnings on to keep an eye on things.

PR #101, opened 2015-12-29
General changes to x11 port (Xv accel, fullscreen, aspect control, joysticks, etc)

greg-kennedy:
Happy holidays :gift: 

These patches backport two features of the GTK branch into the base unix version.  They are:
* Fullscreen-on-startup, and
* 2d hardware acceleration (scaling) through Xv extension.

If you don't have XVideo (or don't want it: configure switch), the changes should be #ifdef'd to properly route around everything.  I've only tested this on one machine, which is old and lacks even an RGB overlay, and verified that it does properly accelerate YUY2 format... so, "works for me".


OV2:
Thanks for the changes. I've only been able to test in a VM so far, where -xvideo produced only a black screen (Fallback to YUY2) - I'll try to set up a real machine if I can find the time.

greg-kennedy:
Thanks, now I get to kick Arch Linux to get them to update their AUR packages : )

PR #119, opened 2016-08-13
Add mute on turbo translation and hack to ensure EVERYTHINK is translated

jristz:
The actual way all is codded make poedit not able to catch ALL the strings therefor not all is translated.

This hack that is just add manyally the entries for every term and words and frasing ensure all is translated when rendered ath the end, the only problem is that we need know which new string is added, but that no bug deal if you have enought time and determination.
Manualy add the strings to snes9x.ui not help since the final rendered program will have duplicates entried or texts.

Also added the new "mute on turbo" translation.

Signed-off-by: Pablo Lezaeta Reyes <prflr88@gmail.com>

### DIFF
--- a/gtk/po/es.po
+++ b/gtk/po/es.po
@@ -7,8 +7,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Snes9X\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2016-07-08 12:52-0400\n"
-"PO-Revision-Date: 2016-07-25 23:59-0400\n"
+"POT-Creation-Date: 2016-08-03 03:09-0400\n"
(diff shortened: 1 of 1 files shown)


bearoso:
You need to run intltool-update to get the extra strings when they aren't in the po. 

The strings in the base snes9x aren't translatable because they have a limited font that gets inserted into the SNES output. Only the gtk UI strings can be translated, so the other directories don't need to be added.

jristz:
I should specify, I'm talking about the GTK UI, like some of the dropdowns menus and rom info page.
using intltool-update left with even less string and more untranslated part and dont bring again all the dropdowns menus not the copyright menu or the credits menu.
so the only way to properly translate all the gtk ui string is doing the search manualy.

PR #133, opened 2016-10-15
Copyright updates

twinaphex:
These are copyright updates on the parts we have created and/or modified in the past -

Evidence of us having touched these parts can be found in these commits -

https://github.com/snes9xgit/snes9x/commit/055aeb5b7a6cfb854e4debc43bcb327f4ac41f68

https://github.com/snes9xgit/snes9x/commit/f33f96872d355814599082e87466e76969991337

### DIFF
--- a/apu/apu.cpp
+++ b/apu/apu.cpp
@@ -148,6 +148,9 @@
   (c) Copyright 1998 - 2001  John Stiles
   (c) Copyright 2001 - 2011  zones
 
+  Sound emulator code compilation fixes
+  (c) Copyright 2013 - 2016  Daniel De Matteis (UNDER NO CIRCUMSTANCE 
+           WILL COMMERCIAL RIGHTS EVER BE APPROPRIATED TO ANY PARTY)
 
   Specific ports contains the works of other authors. See headers in
   individual files.
--- a/libretro/libretro.cpp
+++ b/libretro/libretro.cpp
@@ -1,5 +1,189 @@
-#include "libretro.h"
+/***********************************************************************************
+  Snes9x - Portable Super Nintendo Entertainment System (TM) emulator.
 
+  (c) Copyright 1996 - 2002  Gary Henderson (gary.henderson@ntlworld.com),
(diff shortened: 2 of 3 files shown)


bearoso:
I think a separate comment block in each file would be appropriate if you don't want to be globally in all files. The no commercial part is a bit overboard, too. It's a given that the Snes9x license is no-commercial unless an exception is given by all parties. Unfortunately, I don't think shouting about it in the source code is going to make a difference to the jerks who knowingly appropriate your work for money.

twinaphex:
@bearoso The non-commercial part is not going overboard, as evidenced by certain parties like Nostalgames that tried to 'bribe' certain developers through e-mails, with 'kickstarted' money. It needs to be made clear that shenanigans like this are not going to be going down. Non-commercial needs to mean non-commercial, and since it's pretty much impossible to contact all of the authors, it might as well be made clear from the get-go.

bearoso:
Something more professional, not all caps, would be in order if you're going to make it clear you'll never give up rights. It kind of undermines the fact that otherwise the only all-caps expression in the copyright is "PERSONAL USE." You also don't need to duplicate it on separate lines if you and Themaister have the same copyright dates and policy. I'd do something like:

```
(c) Copyright 2011 - 2016  Hans-Kristian Arntzen,
                           Daniel De Matteis
                           (Under no circumstances will commercial rights be given)
```
And just put it up top under the OV2, BearOso line in all files if you're going to claim copyright for something in the core. If it's everywhere it simplifies updating because I can just use sed to change things. We can do the separate section for the port, too.

twinaphex:
OK, do you want to make the necessary alterations since you seem to have a good idea of how it's supposed to be? Any changes you make I would be OK with in this regard.

bearoso:
Ok, I'll do it.

bearoso:
Hows that look?

twinaphex:
That's awesome, thanks!

We can close this PR now. Thanks a lot!

PR #144, opened 2016-11-11
Introduce --with-system-zip configure flag to use system zip (minizip)

Polynomial-C:
if available.


bearoso:
I can understand the motive behind this and see no reason not to include it.

PR #148, opened 2016-11-15
Add cortex-a53 to libretro Makefile

loganmc10:


loganmc10:
The Raspberry Pi 3 has a Cortex A53 processor so this adds support for that

PR #149, opened 2016-11-15
Add Android (arm) build to libretro makefile

loganmc10:


loganmc10:
this allows the core to be built with the standalone toolchain like this:

```
platform=android make
```

PR #156, opened 2016-12-08
Are you interested in adding the debian directory for deb packages?

qwIvan:


bearoso:
I was told in the past to explicitly NOT do a debian directory by debian maintainers. It was there before and I removed it. Seems they want to do it themselves.

PR #160, opened 2016-12-28
[BS-X] Satellaview Signal Data support

LuigiBlood:
Adds a new folder called "SatData" which contains BSX????-???.bin files, which contains signal data downloaded to the Satellaview base unit registers.
Allows more Satellaview games to work with the right files (should we include such files in there?)

**Only works on Windows for now.**

Unix support could easily be done but I don't have a Linux VM or anything to test with so I didn't make any changes for that.
Mac OS support: I have no clue how that even works.
Libretro support: Seems the folder support isn't really the same so something needs to be done about that.


LuigiBlood:
So I was hoping to talk to devs here for a while and talk about those changes and how should things be done because I'd like to have code everyone could be happy with...

OV2:
Sadly I have almost no experience with Satellaview games and their data. I think we could put the data in the BIOS_DIR, but I also have no problem with the new folder type.
Do different games need different files?

display.h
```diff
@@ -207,7 +207,8 @@ enum s9x_getdirtype
 	IPS_DIR,
 	BIOS_DIR,
 	LOG_DIR,
-	LAST_DIR
+	LAST_DIR,
```
OV2: LAST_DIR should always be the last entry in the list, it is just a marker for the end of the enum.

LuigiBlood: ...I don't even remember why I did this to begin with. I probably did it a little too quick.

LuigiBlood:
I don't want to clutter the BIOS DIR with Satellaview files. And it's not BIOS stuff anyway.
And no, different games shouldn't need different files.

PR #171, opened 2017-02-18
Visual Studio solution modifications

bonimy:
Makes a few changes for building in Windows Visual Studio 2015


OV2:
I'll take a look at merging this in tomorrow. Do you also want to update how2compile.txt?
The platform toolset will have to be switched back to v140_xp, there's no reason not to support winxp atm.

bonimy:
Yeah I'll update the how2compile, but I'll need some time. I need to see what it's like putting all of this in on a fresh install. It'd be nice for other windows users to verify the ease and accuracy of building.

OV2:
I don't think there are that many windows users that actually compile it, but we'll see if you get any feedback ;)

qwertymodo:
Build instructions are more useful to potential contributors. I wouldn't have bothered with MSU-1 support if I couldn't figure out how to build it.

bonimy:
I'm on Windows 7, and one of the biggest issues I run into is DirectX 2008, Windows 8.1 SDK, and XAudio. DirectX and XAudio both have an XAudio2.h include file. If I specify windows as the first include path, the program compiles, but can't run because the SDK's xAudio isn't backwards compatible with Win 7, if I specify the DirectX include first, I break a lot of other things. So I had to include a copy of the DirectX xAudio2.h file in the snes9x solution, but this is specific to windows 7. Any ideas what we should do here?

OV2:
I compile it with the 7.1A SDK (which if compiling for v140_xp is automatically added to the include paths) and I have the dx sdk first in the include paths. That works fine here.

bonimy:
Ah that did the trick. Alright, I'll keep it on v140_xp then. Thanks.

jcdenton2k:
Supporting a deprecated OS that is over 10 years old and has wide-open gaping security holes.  YAY!
:+1:

PR #182, opened 2017-04-01
VS2017 solutions update

bonimy:
PLEASE TEST IN WINDOWS 7 AND WINDOWS XP BEFORE MERGING

Update windows solution files to VS 2017.

This also removes the MASM build commands from the zlib project. Having
this only makes buiding the source harder for new users, and it
ultimately isn't necessary.

The changes also rearrange the snes9x project's <IncludePath> and
<LibraryPath> from (Include);(DirectX);(Cg) to (DirectX);(Cg);(Include).
These changes reflect the original library path order. The modification
I made beforehand could have affected XP builds and I didn't have a
chance to test this before merging.

I also committed the .gitignore file. It removes common files that can show up during the development phase that we do not want in the repository. While it removes library files, I was sure to add exceptions for the library files already in macosx/. No other default files are subject to removal from the .gitignore. This change also makes .gitignore viewable in snes9xw.sln. It also updates the solution to be opened in Visual Studio 2017 by default. This doesn't prevent it from being opened in VS15 for those not ready to make the switch.


OV2:
I haven't gotten around to install VS2017 yet, but I'll try to find some time this weekend.
What was the issue with the masm commands? They should work out of the box and are from the reference zlib project files.

Orphis:
I've tried setting up an automated build for the emulator and I'm hitting issues with masm in zlib. If that change could be extracted in another PR and merged quickly, it would be nice!

OV2:
The masm removal didn't work since ASMV and ASMINF were still defined. I've merged in your changes and removed those.

Orphis:
Well, you do not really need to pull the asm optimized assembly for zlib on Windows at all. I've seen very few projects using them actually.

OV2:
I've removed the asm with this merge, maybe I worded that badly. The defines also needed to be removed so that the build worked correctly.

The default VS projects supplied by zlib include the asm if I remember correctly, that's where it came from.

bonimy:
Okay. I removed the asm because I didn't think masm was a standard feature. I think I mixed it up with another project of mine that uses 'nasm', which users have to download.

Thanks for taking care of that.

PR #183, opened 2017-04-01
Remove DirectDraw

bonimy:
Removes all references to DirectX's deprecated DirectDraw. This should undergo exhaustive testing on all supported windows versions and DirectX SDKs (even the now-deprecated June 2008 SDK). The suggested version of DirectX to use is now the [June 2010 SDK](https://www.microsoft.com/en-us/download/details.aspx?id=6812).

If you merge this, then close #184.


qwertymodo:
Rather than removing it outright, why not make it a conditional compilation with a preprocessor flag?

bonimy:
I originally thought of doing a compile time switch. These were the pros and cons I considered before wanting to go with removing it completely.

Pros:
- Still supporting DirectDraw doesn't remove anything from the snes9x program, allowing users to have the same experience.
- Compile time switches are more flexible.
- @OV2 mentioned we could put ddraw.lib in the repo.

Cons:
- DirectDraw is no longer supported by Microsoft. In fact, the entire June 2008 SDK is no longer available for download from Microsoft.
- If we remove it, then users can build snes9x on any version of DirectX starting at the June 2008 SDK.
- We already have two working alternatives: OpenGL and Direct3D.
- Trying to continue supporting it could be confusing to new contributors. Speaking from personal experience, just trying to build on windows already is a pain.

The cons are admittedly weak points since they're all true with a compile switch. My biggest concern is that, is there really a _reason_ to support DirectDraw? Does it break the functionality of snes9x if we remove it? As far as I know, removing it still lets it run on Windows XP. Personally, I think snes9x would be better off if we kept its dependencies up to date and removed legacy libraries code where we already have modern alternative (FMOD is another example of this). Another idea is replacing DirectDraw with Microsoft's recommended replacements, Direct2D.

I don't want to start a whole debate though, so I'll make a compile switch pull request and the contributors can decide which is a better fit.

qwertymodo:
I really see it as a "if it isn't broke, don't fix it" situation. Like you said, your cons can be equally addressed with a compile-time flag. When you go to add DX11+ support you can change the conditional compilation to something like #if defined(NO_USE_DDRAW) || defined(USE_DIRECTX_11) || defined(USE_DIRECTX_12)

Or if DX defines its own version flags just use those, but you get the idea. Basically just use the ORs to remove ddraw automatically when building against newer versions.

qwertymodo:
FMOD, on the other hand, isn't publicly available anymore (you actually have to sign up and email them to get access to it). That one I feel like should probably be removed entirely. At least MS still publicly hosts old DX packages, even if they're a little hard to find.

bonimy:
I haven't been able to find the 2008 SDK hosted by Microsoft anywhere.

bonimy:
The compile switch version is at #184.

PR #184, opened 2017-04-02
DirectDraw compile switch

bonimy:
This puts all Directdraw-related code into a compile-time switch
'DIRECTDRAW_DEFINED'. To enable DirectDraw, you need to define this in
the snes9xw project file, and add a reference to drraw.lib in it too.

I still kept some DirectDraw-only functions (like SwitchToGDI() and RestoreSNESDisplay()) since there were ambitions of abstracting these functions.

If you merge this, then close #183.

### DIFF
--- a/win32/CDirectDraw.cpp
+++ b/win32/CDirectDraw.cpp
@@ -187,8 +187,7 @@
   Nintendo Co., Limited and its subsidiary companies.
  ***********************************************************************************/
 
-
-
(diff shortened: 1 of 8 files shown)


OV2:
I've merged this with some changes (the enum remains the same for compatibility etc.). At some point we might remove it entirely, for now I'll keep it in my compiles.

PR #190, opened 2017-04-25
GTK: Add a 10th save slot (Slot 9).

YamashitaRen:
... Reorganizing save slots section of snes9x.ui while I'm at it.

I was annoyed at snes9x-gtk having save slots only up to 8 when other emus and even windows port have save slots up to 9.

This pull request is a step towards feature parity.


bearoso:
This seems kind of superfluous, but I'll go along with it just for parity with the other ports.

PR #191, opened 2017-04-25
Remove CG dependency on Windows

Orphis:
CG is old technology and deprecated in favor of better APIs.


bearoso:
If you're not going to add support for those better APIs, don't post a pull request removing a widely-used function. Making the build easier is not worth losing this feature.

qwertymodo:
Just because it's deprecated doesn't mean it should be removed, at least not until we have something better to replace it with.  If somebody wants to implement GLSL shaders, then I could see removing CG, but there is nothing to gain by removing it now.

Orphis:
Well, you have HLSL shaders instead. An old deprecated technology at this point is just tech debt.

bearoso:
> Well, you have HLSL shaders instead. An old deprecated technology at this point is just tech debt.
Where do you have HLSL shaders? The only alternative implemented is the rudimentary ruby 1.0 glsl format. Removing Cg removes all multipass shaders and the capability completely from the Direct3D output.

Orphis:
Display Settings > D3D shader file or OpenGL shader file.
By keeping Cg aroung, you're just asking people to keep developing new shaders with it instead of using better solutions. This is a development build anyway, not a release, so people would have time to adapt.
And I'm also pretty sure that most effects people implemented in Cg are also available in other languages.

Orphis:
Checked the multipass feature, and there are people doing the same thing with GLSL in RetroArch, and they even have a script to convert Cg shaders to GLSL. It could be a way forward.

jcdenton2k:
It would help if @bearoso and @qwertymodo were willing to work with @Orphis and the RetroArch project to share code and improve this vanilla version of SNES9X instead of just making excuses and saying no.

I'm happy that Orphis is willing to work to improve things.  :)

Orphis:
I didn't say I was going to do it though. Since there's barely any documentation, it wasn't obvious why the Cg shaders were required, and even less that it was the only way to currently support multipass shaders.
We'll see if I find the time and interest in doing so.

bearoso:
I'm only in charge of the Gtk+ branch, I don't know anything about the Windows port. I was merely pointing out why removing Cg would alienate a lot of users.

That said, I've been meaning to try and add the glslp format to the Gtk port eventually, since I never got around to adding Cg.

qwertymodo:
I'm not a maintainer, I spemt literally one week implementing MSU-1 and that's the extent of my involvement. I only show up to talk here because now I get notifications of activity.

OV2:
As stated before simply removing it is not an option. However I've committed a change so that you do not have to have the cg toolkit installed to compile it.

OV2:
Build dependency has been removed in 76c2d05187b870bd6dd1b345e2d611066b7520c5

PR #196, opened 2017-05-10
Add xinerama multi-monitor support and position window hints

Francesco149:
This adds support for multi-monitor xinerama setups.

The desired monitor is selected through -xineramahead or ```XineramaHead``` in ```[Unix/X11]```.

Also, I've added position hints for the window, as many window managers (including mine) ignore x/y provided to ```XCreateWindow``` .

### DIFF
--- a/unix/configure
+++ b/unix/configure
@@ -719,6 +719,7 @@ enable_jma
 enable_screenshot
 with_x
 enable_xvideo
+enable_xinerama
 enable_sound
 '
       ac_precious_vars='build_alias
@@ -1370,6 +1371,7 @@ Optional Features:
   --enable-screenshot     enable screenshot support through libpng (default:
                           yes)
   --enable-xvideo         enable Xvideo if available (default: yes)
+  --enable-xinerama       enable Xinerama if available (default: yes)
   --enable-sound          enable sound if available (default: yes)
 
 Optional Packages:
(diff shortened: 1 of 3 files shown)


qwertymodo:
You should go back and rebase this commit to remove all of the version number changes.  Also, not sure why you removed $LDFLGS here https://github.com/snes9xgit/snes9x/pull/196/commits/b3e2750eac3bea9f581d3fe0c83c1e6ccabdb2d2#diff-425472bfd346dd18d9d71bd323873e15L6311

Francesco149:
I simply ran autoconf to regenerate the configure script after adding the xvideo configure flag, is that incorrect? The versions numbers in there are just autogenerated by whatever version of autoconf you're using, and configure.ac is what dictates that output.

qwertymodo:
It's correct, but you should have done a partial commit on unix/configure using git add --patch or git add --interactive and only included the relevant code in the commit here.

bearoso:
It's ok, I'll look over the patch later.

Francesco149:
ah, my bad, it's just that the included configure script was outdated, I guess (it was 1.54 before). I can see why you would want a clean commit though. I'll see if I can fix it in the meantime.

Francesco149:
should be ok now

Francesco149:
fixed a minor mistake in how invalid xinerama heads are handled.

marcosfrm:
@qwertymodo I removed it here:

https://github.com/snes9xgit/snes9x/commit/359b906b69f0443c02942fe6ac06513607cfeb7a

It was being passed when compiling objects and not on linking.

PR #200, opened 2017-05-19
Silence -Wimplicit-fallthrough= warnings

orbea:
Silences countless `Wimplicit-fallthrough=` warnings with gcc-7.1.0.
```
../tile.cpp:904:41: warning: this statement may fall through [-Wimplicit-fallthrough=]
     case 6: DRAW_PIXEL(6, Pix = bp[6]); if (!--w) break; \n                                         ^
../tile.cpp:1475:2: note: in expansion of macro ‘DRAW_TILE’
  DRAW_TILE();
  ^~~~~~~~~
../tile.cpp:905:5: note: here
     case 7: DRAW_PIXEL(7, Pix = bp[7]); break; \n     ^
../tile.cpp:1475:2: note: in expansion of macro ‘DRAW_TILE’
  DRAW_TILE();
  ^~~~~~~~~
```
This was backported from snes9x2010, see these two links.
https://github.com/libretro/snes9x2010/blob/master/src/tile.c#L927
https://github.com/libretro/snes9x/pull/48

### DIFF
--- a/tile.cpp
+++ b/tile.cpp
@@ -843,7 +843,7 @@ void S9xSelectTileConverter (int depth, bool8 hires, bool8 sub, bool8 mosaic)
 #define DRAW_TILE() \
 	uint8			*pCache; \
 	register int32	l; \
-	register uint8	*bp, Pix, w; \
+	register uint8	*bp, Pix, w, n; \
 	\
 	GET_CACHED_TILE(); \
 	if (IS_BLANK_TILE()) \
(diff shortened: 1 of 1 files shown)


orbea:
No comment?

OV2:
I'll merge it in when I find the time for some performance checks.

orbea:
Cool, thanks for letting me know . :)

bearoso:
I'm seeing a 7% total speed loss with this changed, so the optimization trick is making a huge difference.

qwertymodo:
The correct solution is to leave it as is and just use the [[fallthrough]] attribute, but I don't know if that breaks older compilers, you might need a conditional.

orbea:
Well, the fallthrough attribute is not desirable, the volume of warnings makes any other compiler output useless and suggests it may break with future gcc versions. Though I I can understand that a speed loss is not desirable either. Do you have any other suggestions?

bearoso:
Fixed this using comments as mentioned in the gcc docs.

orbea:
Thanks for the time, I will try to make the changes on the libretro core later too.

PR #208, opened 2017-06-28
Update libretro.cpp

loganmc10:
Taken from https://github.com/libretro/snes9x

This was just a copy and paste from the libretro repo, I just took out the SuperFX overclocking bits, since they aren't supported here.

Tested a few games, works as normal.

### DIFF
--- a/libretro/libretro.cpp
+++ b/libretro/libretro.cpp
@@ -9,34 +9,40 @@
 #include "snapshot.h"
 #include "controls.h"
 #include "cheats.h"
-#include "movie.h"
 #include "logger.h"
 #include "display.h"
 #include "conffile.h"
 #include <stdio.h>
-#ifndef __WIN32__
+#ifdef _WIN32
+#include <direct.h>
+#else
 #include <unistd.h>
 #endif
 #include <sys/stat.h>
 #include <sys/types.h>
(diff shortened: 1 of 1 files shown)


loganmc10:
My basic thinking here: the libretro.cpp in this retro isn't really maintained by anyone, but it is looked after in the libretro repo. Copying it over and removing the SuperFX overclocking is a very easy way to keep it up to date. There might even be some functions in this libretro.cpp that end up being "dead code" (not used anywhere), but they will get removed by the compiler if they are dead code, so I think this is better than having an unmaintained libretro.cpp

OV2:
The one in this repo is maintained by me, I usually make sure that it builds and can at least run roms.
I'll take in the new stuff from this PR and throw out the things not working in mainline (mem mapping, achievements) and see if the aspect ratio stuff makes sense. It also doesn't build at the moment since the movie function is missing.

loganmc10:
hm it built for me, anyway thanks for looking at it

OV2:
I've finally merged the changes today, sorry it took so long. I've also re-added the 8:7 AR - the NTSC AR that was supposed to be the same looked more like 4:3.

loganmc10:
Thanks!

Issue #38, opened 2013-08-15
[Question] How contributr to translating??


jristz:
So that...
I want contribute to translating the GUI to another language

so you have transifex, or any fancy mathematical place for that or like the old way of clone and pull??

vgturtle127:
We do not have a transifex, and you should probably tell us what language you want to translate snes9x into first, since it may already be in that language.

Language isn't about math, by the way? :laughing:

jristz:
Spanish (es), i have experience translating Xfce and Voxelands, both use
transifex, but also i know how poedit work.

vgturtle127:
SNES9X doesn't use .po files either. I think translating might be an issue, since I believe everything is hardcoded...

OV2:
The gtk version supports translations via po files, the windows version is not translatable without rebuilding.

vgturtle127:
Oh, I didn't see that. Is the GTK version going to be upgrading to GTK3 soon? I only ask because GTK3 seems to cause some issues when building, and if we wanted people to help translate that version then we would want to upgrade first before anybody started modifying the current files.

Also, is there a chance we may move away from GTK, or use the GTK version on Windows or something?

jristz:
Gtk2 is near death in the sence tha maybe no more updates will be shipped,
unless snes9x want get frozen in Gtk2 and risk Gtk2 bugs then better
migrate to Gtk3 Lije Xfce is doing or QT5 like Lxde, not mather what the
end I want translate Snes9x.

vgturtle127:
@jristz Haha, OK. I can see you are interested. Is the current https://github.com/snes9xgit/snes9x/blob/master/gtk/po/es_VE.po file any use, or is it a different regional version of Spanish?

jristz:
El lun 28/09/2015, 1:56 PM, vgturtle127 <notifications
<notifications@github.com>@ <notifications@github.com>github.com
<notifications@github.com>> escribió:

@ <https://github.com/jristz>jristz <https://github.com/jristz> Haha, OK. I
can see you are interested. Is the current https
<https://github.com/snes9xgit/snes9x/blob/master/gtk/po/es_VE.po>://
<https://github.com/snes9xgit/snes9x/blob/master/gtk/po/es_VE.po>github.com
<https://github.com/snes9xgit/snes9x/blob/master/gtk/po/es_VE.po>/
<https://github.com/snes9xgit/snes9x/blob/master/gtk/po/es_VE.po>snes9xgit
[...]

PR #223, opened 2017-08-26
Minor update to es.po

jristz:
Minor update to spanish translation so Slots in contectual menu get translated (I hope)

Signed-off-by: Pablo Lezaeta Reyes <prflr88@gmail.com>


bonimy:
Does this close #38 or is there still more work to do?

jristz:
No, is just an update for the 9th save slot string. 

The problem is that the strings arent caught by any tool making the only option to manualy update on every New string

PR #236, opened 2017-09-09
GTK/unix fixes

megari:
This pull request contains fixes to make the GTK and unix frontends build and work again after the merge of the msu1-zip branch, as well as fixes to some issues detected by g++ while I was debugging the original issues.

The last patch is the least important, as the issues are minor nitpicks, but the first two fix serious issues.

### DIFF
--- a/apu/bapu/smp/core.cpp
+++ b/apu/bapu/smp/core.cpp
@@ -51,8 +51,8 @@ void SMP::op_write(uint16 addr, uint8 data) {
 
 void SMP::op_step() {
   #define op_readpc() op_read(regs.pc++)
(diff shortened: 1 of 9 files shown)


megari:
It seems recent changes made this pull request unmergeable. Will rebase in a jiffy.

megari:
Rebased on current master.

megari:
Found an additional case of undefined behavior, so added the fix for that in the appropriate commit.

OV2:
I've left out the splitpath modifications and changed the call to pass in all parameters, since this is the only location where it was called this way. Otherwise we would have to change all _splitpath implementations: there is also mac and libretro, and possibly in other ports.

PR #238, opened 2017-09-12
Move away from deprecated functions in the GTK port

megari:
Looking at the build warnings, there were two uses of deprecated GTK/GDK functions. These looked easy to replace with non-deprecated ones, so I did so.


megari:
Please note that enabling GTK3 causes a lot more warnings about deprecated stuff to be emitted.

megari:
I now noticed that someone had submitted a pull request to get rid of the use of the deprecated GTK stock items, but this was not deemed acceptable, as this would have degraded the GTK2 version. Would a version of such changes be accepted where the GTK2 version won't be degraded?

In any case, this pull request only contains changes that are compatible with both GTK2 and GTK3.

PR #244, opened 2017-10-07
Apu code includes

### LINKED ISSUE #243: Some APU code files not visible in Visual Studio.
## Issue
In Visual Studio, under the APU folder, there are some files (header and source) not included in the solution explorer. Here is what we see so far:

slnexplorer

However, when looking in the file explorer, we see some files not referenced.

explorer

These files are explicitly included in the source code:

includes

But I'm not sure this is the preferred way to add .cpp files. Especially since they aren't visible in the Visual Studio solution explorer. Further, including code this way removes certain Visual Studio features, namely Intellisense and syntax highlighting. Instead, we see this:

algorithms

The APU file include structure also suffers from other include file errors, so that some files that are included, aren't properly highlighted.

sdsp

## Resolution
Refactor the preprocessor code to a format that Visual Studio likes, without changing any of the functional code.

## Proposed tags
- windows
- documentation
- emulation-apu

bonimy:
## Description
Refactors preprocessor codes in APU source files without changing code functionality.

Add missing APU files to Visual Studio solution explorer.

Closes #243


OV2:
This will break all the other ports, and we gain some code highlighting from VS.

@bearoso any thoughts? Should we keep the APU code as close to the bsnes original, or do we no longer care since there is no performance core in higan anymore?

bonimy:
@OV2, I didn't get to look so much into other ports, so I'm just curious what part of it actually causes the breaking?

bearoso:
I kind of like how it was organized before, but if this makes it significantly easier to work with on Windows, then it’s your prerogative.

The opcode inlining, however, needs to stay. It does make a difference with gcc.

qwertymodo:
Rather than inlining the .cpp files, why not just add new header files to expose those functions to each other across compilation units and include those instead of including the .cpp's?

bonimy:
@bearoso, I can put it back into inline.

@qwertymodo I'll see what I can do with that.

bearoso:
We should probably do this now, but obviously this pull request is now dated.

bonimy:
I've been working on an updated version. If I ever stop being lazy and finish it, I'll PR again.

Issue #22, opened 2012-05-14
Mac: Build Issues


speedofmac:
I'm getting quite a few compilation warnings and errors when compiling snes9x on OS X 10.6.8. The majority of problems seem to stem from the /apu directory. Here's a link to the compilation output: http://pastebin.com/ZJCY29JE

One thing that fixed the majority of issues was to apu/apu.cpp:185 to:

```
#include "bapu/snes/snes.hpp"
```

I'm not sure about the others. Adding the /apu/bapu directory to the project didn't seem to fix the missing file issues, but someone more knowledgeable than I should be able to figure out what's going on.

vgturtle127:
Any updates on this? Mac users haven't been having issues all this time have they?

eisnerguy1:
I still can't compile the master code in Xcode 8.3.3.

Issue #216, opened 2017-08-07
[MacOS 10.12] File->Open ROM Image doesn't open dialog in Sierra


jacobthehammer:
(This is using the 1.5.3 build from 2012, I couldn't find a more recent package for Mac.)

Steps for recreating the bug are fairly simple.
1. Open Snes9x.app
2. Either click File->Open Rom Image or use cmd+o
There is no Finder dialog as expected. 

Using Console in MacOS we can see exactly what happened. I've included a [screenshot here for reference.
](http://imgur.com/a/NmDdg)

![screen shot 2017-08-06 at 8 48 50 pm](https://user-images.githubusercontent.com/9625866/29009498-5f1d35c4-7ae9-11e7-81f4-2d916444104f.png)

`[WARNING] could not find container for id: com.apple.Documents`
`[WARNING] could not find container for id: com.apple.Documents`
`[WARNING] could not find container for id: com.apple.Documents`
`-[NSView setNeedsDisplayOnHICocoaView]: unrecognized selector sent to instance 0x1c254d0`
`Connection to sharingd became invalid`
`Exception raised!: -[NSView setNeedsDisplayOnHICocoaView]: unrecognized selector sent to instance 0x1c254d0`

---

Based on the data from console, I'd chalk this up to some change in Cocoa that Apple has made since OSX Lion came out in 2012. 

Thanks for making such a great emulator!

OV2:
Unfortunately there is no maintainer for the osx port at the moment, so this is unlikely to get fixed anytime soon.

GB-CG:
@jacobthehammer There's the 1.54.1 port on OpenEmu, but that's another story.

mcclure:
The error triggers somewhere inside of Apple's `NavDialogRun` function (called in `NavOpenROMImage`).

`NavDialogRun` is a VERY old system function (it does not even exist in 64-bit executables) and I think it just stopped working correctly with Sierra…

The entire open/save dialog system in snes9x seems to be written for compatibility with Mac OS 9 and so it uses a lot of functions that are deprecated…

bearoso:
Apparently fixed.

PR #256, opened 2017-10-29
recent macOS compatibility

tmkk:
This will fix issue #22 and #216.
It builds successfully with Xcode 8, but 10.6 SDK is required for deprecated functions/constants.


tmkk:
Now works with up to 10.11 SDK. 10.12 SDK is still incompatible because QuickTime headers are missing.

OV2:
We don't really have a way to compile or test the mac port, so I'll just merge it.

PR #259, opened 2017-11-12
unix: fix fullscreen property change timing

NoSuck:
Previously, the Unix version of Snes9x was unable to create a fullscreen window.  (It would create a regular window the size of the screen.)

### DIFF
--- a/unix/x11.cpp
+++ b/unix/x11.cpp
@@ -862,6 +862,12 @@ void S9xInitDisplay (int argc, char **argv)
 							WidthOfScreen(GUI.screen), HeightOfScreen(GUI.screen), 0,
 							GUI.depth, InputOutput, GUI.visual, CWBackPixel | CWColormap, &attrib);
 
+		/* Try to tell the Window Manager not to decorate this window. */
+		Atom wm_state   = XInternAtom (GUI.display, "_NET_WM_STATE", true );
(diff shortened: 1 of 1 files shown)


greg-kennedy:
Why did you make this change?  Seems like it would make it unable to create a NON-fullscreen Unix X11 window now.

greg-kennedy:
My bad.  Diff context cuts off right before the `if (GUI.fullscreen == true) {` line, so I didn't see that it was still wrapped by conditional :)

greg-kennedy:
Sorry to pollute this issue with more chatter, but... I did actually test this on my setup, and this change did break fullscreen handling for me.  (Non-fullscreen works fine: as you pointed out, I'm stupid lol)   With this change I'm having the same result now as @NoSuck had before!  So it fixed it for him, and broke it for me.

The problem centers around when you are supposed to call XChangeProperty, and I am pretty sure it is dependent on your window manager.  Before this change, snes9x waits until after XMapRaised before making the property change to switch into fullscreen.  This worked for me using my wm, `lwm`.

But this change causes snes9x to call XChangeProperty  *before* XMapRaised (but immediately after XCreateWindow).  `lwm` doesn't like that but I'm guessing others do.  I'm interested to know what window manager you both are using.

X WM coding is a black art and I think this is undefined behavior anyway, so I'm gonna just leave this comment here as a warning to anyone else who runs into it: maybe go fix your WM to catch atoms before XRaiseWindow, or maybe revert this patch manually when you compile (that's what I did).

PR #266, opened 2017-12-16
BS-X Cleanups and Fixes

LuigiBlood:
Just done a lot of QoL fixes, cleanups, and more accuracy.

- Avoid updating the memory map when it is not needed
Basically using the dirty variable again as it is supposed to.

- Support for bigger SatData files
When I was making SatellaWave for making those files, I found out I did the implementation wrong; as it would have been, it could only download like up to 300KB of data, when you have videos proving the contrary... This makes it support bigger files.

- BS-X Memory Mapping fixes
Cleanups
Proper PSRAM mapping (forgotten mapping and "over"mapping fix)
Flash Interface not limited to bank 0xC0
(MMC would not interfere because of its particular addressing)

- Flash Support for BS Compatible Games
Removed write MMC check from S9xSetBSX() [redundant with BSX_Map_FlashIO()]
(Also basically I forgot why I added particular lines and found out why I did that, so this time, I commented why)

### DIFF
--- a/bsx.cpp
+++ b/bsx.cpp
@@ -675,8 +675,6 @@ static void BSX_Map (void)
 
 	memcpy(BSX.prevMMC, BSX.MMC, sizeof(BSX.MMC));
 
-	printf("BSX MAP\n");
-
 	MapROM = FlashROM;
 	FlashSize = FLASH_SIZE;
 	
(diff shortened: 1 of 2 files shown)


LuigiBlood:
So there's one other thing I should fix.

The downloading actually goes so fast, the BS-X doesn't even have time to properly do things.
I even got unstable music because of it. I should slow it down a little bit.

clobber:
Just pinging @bearoso and @OV2 to see if they are ready to merge.

PR #283, opened 2018-04-28
Cheats Restructure

bearoso:
This branch restructures the cheats support. Descriptions are longer, multiple codes can be grouped, we use the same file format as higan--converting old .cht files if we find them, we include the mightymo cheat database, and the issues with saved bytes and applying should be mostly fixed.

For Gtk+, I added the ability to look up the current game in the included mightymo database (data/cheats.bml). It matches the now-loaded SHA256 sum, then looks it up in the database. A possible future fallback might be to also compare the rom name or internal name, or allow the user to search for a string.

For Unix, cheats are mostly broken and no one noticed. Any specified on the command line are erased when the game image tries to load the .cht file. It's still that way, for S9xLoadCheatFile compatibility's sake.

For libretro, cheats are reloaded on any change, and they don't use the .cht file, so we're mostly good. Goldfinger support wasn't done correctly and never worked. No one noticed, so I removed it.

For win32, I removed the address and byte dialog boxes from the cheat window, as they make no sense with multiple codes in a string. The cheat search's add cheat dialog might have some vestigial entries, but seems to work. I didn't add the database lookup because I was unfamiliar with where the Win32 port likes to keep stuff. Goldfinger support was missing here, and now it's gone completely except for the conversion function in cheats.cpp. I suspect no one will miss it.

@OV2, would you review the win32 stuff and make sure it's working on your end?


OV2:
Some things I noticed:
- S9xLoadCheatFile is called in the LoadROM functions, but S9xDeleteCheats is only called when it loads classic cheats, so cheats from a previous game will still be active. In the GTK port you call that after loading the rom. Any reason we should not also call it in the LoadROM functions?
- You moved S9xCheatsEnable to the ports - any particular reason for that?
- The new format still uses the cht extension, should we change it to prevent confusion with older versions?

The win32 changes look fine, I'll see that I can add the database lookup soon.

bearoso:
The S9xEnable/DisableX is a replacement to S9xApplyCheats and the Settings.ApplyCheats flag. This is because the previous way was ambiguous. You had to set the enable flag on the cheat and run S9xApplyCheats to enable it. But does that set the Settings.ApplyCheats flag or use it to determine whether to apply them? Then S9xApplyCheats was also used to update the state in memory. It was confusing. So now Settings.ApplyCheats is a port setting, not used internally. S9xCheatsEnable is the global toggle, like for the shortcut key in the Windows port. In Gtk, I just call it at initialization and forget it. 

S9xLoadCheatFile is supposed to delete existing cheats no matter the format. I must have put that S9xDeleteCheats back in the wrong function after experimenting with the Unix port. I do think it would be better in another location. I’ll think about it.

bearoso:
I’m not sure about the .cht extension. Higan uses a cheats.bml in the cartridge folder, so that’s out for single-file roms. I like the fact that on import the new format writes over the old file so there’s no confusion as to which is being loaded.

qwertymodo:
>The new format still uses the cht extension, should we change it to prevent confusion with older versions?

I think we should change this.  We could simply use gamename.bml, since that's the actual file format, but .bml can contain things other than cheats (such as full memory maps), but I think that shouldn't be a problem for us since the cheat files live within their own directory, so there isn't really any ambiguity for CheatDir\gamename.bml.  It's not currently a problem anyway since we're not using bml files for anything else, but I figured it was at least worth mentioning.

bearoso:
> but I think that shouldn't be a problem for us since the cheat files live within their own directory, so there isn't really any ambiguity for CheatDir\gamename.bml. It's not currently a problem anyway since we're not using bml files for anything else, but I figured it was at least worth mentioning.

You're partially wrong on both counts. Cheats don't always live in their own directory--it's configurable. On the Gtk+ port, at least, it's the same directory as the ROM file by default, and I know that's at least an option on the currently decrepit OS X port. Plus, we may start using bml files for more things in the future. But I don't think those future uses would conflict, so .bml is fine if you guys think it is. 

I fixed S9xLoadCheatFile so it doesn't delete the current list of cheats, but the LoadROM functions do. I've got the unix port working like it used to, with the exception that if you pass -cheat to load the cheat file and -gamegenie or friends on the command line, they're both used. It's still destructive, saving whatever it had loaded to the cheat file or deleting it, no matter what.

bearoso:
Ok, using .bml now. We still try to load any .cht, even if they're .bml format, but only save to the .bml.

OV2:
I've added cheat lookup to the win version and did some fixups in the dialog. When I transferred the search_database function I noticed that loading from CHEAT_DIR and ROM_DIR will not work as intended because S9xGetFilename inserts the rom filename.

If we use .bml as the extension we should probably ask byuu, we don't want to interfere with any possible cartridge folder usage.

bearoso:
> I've added cheat lookup to the win version and did some fixups in the dialog. When I transferred the search_database function I noticed that loading from CHEAT_DIR and ROM_DIR will not work as intended because S9xGetFilename inserts the rom filename.
If we use .bml as the extension we should probably ask byuu, we don't want to interfere with any possible cartridge folder usage.
Good catch. I should have used S9xGetDirectory and appended cheats.bml. In fact, I thought about that and when I left and came back to it I didn't change it.
Yeah, we'd need to ask byuu about the .bml extension. After thinking about it, I'm not sure it's the best idea, either. Any other suggestions, besides just going back to .cht again?

qwertymodo:
As far as avoiding conflicts with cartridge folders goes, higan uses cartridge.sfc\cheats.bml and the only other .bml file I know of is cartridge.sfc\manifest.bml.  Everything is referenced by full filename, so using the file extension isn't going to cause any issues unless you do something really silly like save the cheats as manifest.bml.  The file extension simply refers to the format, not the contents, so it makes sense to continue using it.

Issue #291, opened 2018-05-09
Wayland support


remicalixte:
Hi !
I am currently trying to make the gtk port running natively on wayland. That is still a very early work in progress but I think it could be cool ! If anyone is interested, you check it at https://github.com/remicalixte/snes9x on the wayland branch.

bearoso:
Working for the most part now. Using OpenGL instead of software driver is highly recommended.

PR #292, opened 2018-05-11
WIP: Wayland support

remicalixte:
Hi!
As is said in #291, I have begun to make the gtk port working natively under wayland.
So far I managed to run games successfully with the software renderer, I don't have much experience with OpenGL so I have not really look this part yet, I have just seen that it uses GLX so there is some work to do.

The easy part was to wrap the X specific stuff in compile-time and run-time checks so that snes9x does not seg fault at launch.
The most important change I made is to send a "draw" signal when S9xDeinitUpdate is called and handle the rendering in the expose callback, because for some reason directly rendering in S9xDeinitUpdate does not work under wayland, even it is working correctly under X11. I tried to do this with the fewest possible changes to the code base.

Is the Wayland support something the snes9x team would like to add ?
I think it could be very useful for several reasons:
 - Wayland is supposed to replace X11 at some point
 - in native Wayland sessions (like Gnome that I use) Wayland applications are rendered nicer than X applications that have to use XWayland
 - Wayland is supposed to be lightweight and performing better than X in less powerful systems like ARM boards (Snes9x is quite popular on this kind of platform)

Moreover, as a side effect, the work I have done makes the gtk port somewhat backend-agnostic and I have successfully ran it on macOS! Considering the macosx port seems unmaintained at the moment and the most recent build I found is a 32-bit one from 2011, I find this interesting (even if gtk on macOS is not as nice as a native application).

Feel free to comment this patches, and if you guys are interested in making it happening I will be very happy to continue working on it!

### DIFF
--- a/gtk/src/gtk_config.cpp
+++ b/gtk/src/gtk_config.cpp
@@ -5,7 +5,9 @@
 #include <libxml/parser.h>
 #include <libxml/tree.h>
 #include <libxml/xmlwriter.h>
+#ifdef GDK_WINDOWING_X11
 #include <gdk/gdkx.h>
+#endif
 #include <gtk/gtk.h>
 #include <gdk/gdk.h>
 #include <gdk/gdkkeysyms.h>
--- a/gtk/src/gtk_display.cpp
+++ b/gtk/src/gtk_display.cpp
@@ -1,5 +1,7 @@
 #include <gdk/gdk.h>
+#ifdef GDK_WINDOWING_X11
 #include <gdk/gdkx.h>
+#endif
 #include <sched.h>
 
 #include "gtk_s9x.h"
--- a/gtk/src/gtk_display.cpp
+++ b/gtk/src/gtk_display.cpp
@@ -10,10 +10,14 @@
 #include "gtk_display_driver_gtk.h"
 #include "snes_ntsc.h"
 #ifdef USE_XV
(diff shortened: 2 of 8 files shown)


bearoso:
Most of this looks fine, but the drawing thing I'd like to avoid. Using the expose signal to update things means that the toolkit is in charge of when things get updated, and, from experience, that's not usually on a smooth basis.

remicalixte:
I understand, actually I noticed that on turbo, the fps went lower with this patch, so maybe using the signal induces a bit of overhead. I will try to find another way to do this. Do you know why this is not working under Wayland ? I have not much experience with gtk.

bearoso:
What we're probably going to do for Wayland is bind the wl_surface of the drawing area to an EGL window and use that to render. Since we have EGL automatically, there's no reason to use the slow software blitting at all.

remicalixte:
This seems like a good idea, and hopefully we can bypass the toolkit using EGL directly

remicalixte:
Great, I guess I can close this PR now. I'm glad my stuff could help a bit. I've tried to add egl support for wayland but I was struggling with gtk/wayland interaction, and the documentation is not really that helpful. I will look at what you've done to learn about this

bearoso:
Closing this now. Any further Wayland issues can be their own bugs.

PR #327, opened 2018-06-07
Backport Hermite Resampler bypass code from Libretro version

Dwedit:
I had already submitted this to the Libretro version, I think I'll do the same here.

This is a change to hermite resampler code to make it operate in "bypass mode" if the ratio is exactly 1, or in other words, the output sampling rate is exactly equal to the input sampling rate.

If the hermite resampler is bypassed, this eliminates the hidden internal state from the audio system, so a savestate will contain all the information needed to generate the exact same samples on the next frame.
It also doesn't need to do any math on the sound samples either, for slightly better performance.

### DIFF
--- a/apu/hermite_resampler.h
+++ b/apu/hermite_resampler.h
@@ -66,6 +66,13 @@ class HermiteResampler : public Resampler
         void
         read (short *data, int num_samples)
         {
+            //If we are outputting the exact same ratio as the input, pull directly from the input buffer
+            if (r_step == 1.0)
+            {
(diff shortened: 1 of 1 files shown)


bearoso:
Keep in mind that the MSU1 is 44100kHz, so when that’s in use a resampler is needed for either the internal or MSU1 audio.

Dwedit:
Thanks for the heads up, I think I might see what would need to be done to add the resampler state to the save states.

PR #351, opened 2018-07-01
Bring the core up to speed

fr500:
I standarized indentation for staters, there were at least four different styles. I backported upstreamable changes like multicart loading and hires blending.

There are other non upstream friendly options in a branch, I'll send a PR for those too so maybe they can be reimplemented in upstream.

The idea of this rebase is to reduce the patch size between upstream libretro, and our libretro.

Let me know if there is anything you think that shouldn't be merged here


libretro/Makefile
```diff
@@ -56,7 +56,7 @@ ifneq (,$(findstring unix,$(platform)))
    CC = gcc
    SHARED := -shared -z defs
    else
-   SHARED := -shared -Wl,--version-script=link.T -Wl,-z,defs
+   SHARED := -shared -Wl,--version-script=link.T
```
OV2: Without this, unresolved symbols will not be caught at compile time. Can we detect solaris in the makefile and remove it only in that case?

libretro/libretro.cpp
```diff
@@ -198,7 +200,16 @@ static void update_variables(void)
...
+
+   var.key = "snes9x_hires_blend";
+   var.value = NULL;
+
+   if (environ_cb(RETRO_ENVIRONMENT_GET_VARIABLE, &var))
+      hires_blend = !strcmp(var.value, "disabled") ? false : true;
+   else
+      hires_blend = false;
+
+   var.key = "snes9x_overclock_superfx";
```
OV2: I think this needs to match the name above. Which one is used in your core? With _superfx added?

libretro/libretro.cpp
```diff
@@ -1076,12 +1113,38 @@ bool8 S9xDeinitUpdate(int width, int height)
...
+	 }
+
+
    video_cb(GFX.Screen, width, height, GFX.Pitch);
    return TRUE;
 }
 
 bool8 S9xContinueUpdate(int width, int height)
 {
+   S9xDeinitUpdate(width, height);
```
OV2: Doing this in ContinueUpdate will update with only half the scanlines ready. Previously that produced corrupted images for example in ASP. That no longer happens in the latest version, so I guess we can take this in, but it still might produce unwanted effects in some games.

libretro/libretro.cpp
```diff
@@ -579,13 +762,29 @@ bool retro_load_game(const struct retro_game_info *game)
...
       {
-         extract_basename(g_basename,   game->path, sizeof(g_basename));
+         extract_basename(g_basename, game->path, sizeof(g_basename));
          extract_directory(g_rom_dir, game->path, sizeof(g_rom_dir));
       }
 
-      rom_loaded = Memory.LoadROMMem((const uint8_t*)game->data ,game->size);
+      if (is_SufamiTurbo_Cart((uint8 *) game->data, game->size)) {
+         if (rom_loaded = LoadBIOS(biosrom,"STBIOS.bin",0x40000))
+            rom_loaded = Memory.LoadMultiCartMem((const uint8_t*)game->data, game->size, 0, 0, biosrom, 0x40000);
```
OV2: I don't think this is actually necessary since the core can automatically load the bios, but we can take it in.

libretro/libretro.cpp
```diff
@@ -602,62 +801,105 @@ bool retro_load_game(const struct retro_game_info *game)
    if (!rom_loaded && log_cb)
       log_cb(RETRO_LOG_ERROR, "[libretro]: Rom loading failed...\n");
 
+   struct retro_memory_map map={ memorydesc+MAX_MAPS-memorydesc_c, memorydesc_c };
+   if (rom_loaded) environ_cb(RETRO_ENVIRONMENT_SET_MEMORY_MAPS, &map);
```
OV2: We don't support the memory mapping feedback, but I guess we can leave the callback and functions in if it makes it simpler for you.

libretro/Makefile
fr500: I'll ask the @kwyxz

libretro/libretro.cpp
```diff
@@ -678,6 +920,22 @@ void retro_init(void)
...
+      snprintf(retro_system_directory, sizeof(retro_system_directory), "%s", ".");
+
+   if (environ_cb(RETRO_ENVIRONMENT_GET_SAVE_DIRECTORY, &dir) && dir)
+      snprintf(retro_save_directory, sizeof(retro_save_directory), "%s", dir);
+   else
+      snprintf(retro_save_directory, sizeof(retro_save_directory), "%s", ".");
+
+   // State that SNES9X supports achievements.
+   bool achievements = true;
+   environ_cb(RETRO_ENVIRONMENT_SET_SUPPORT_ACHIEVEMENTS, &achievements);
```
OV2: Since we don't have the memory mapping feedback I think we also need to set this to false, right?

libretro/libretro.cpp
fr500: ahh true, between patches I missed that one, I'll fix it tomorrow

libretro/libretro.cpp
fr500: not sure what you mean with memory mapping feedback, achievements are read only (we only read memory)

fr500:
I can ask again about memory maps, @kps501 and @leiradel are more knowledgeable on that, I just backported everything since libretro/snes9x was getting too difficult to mantain

libretro/Makefile
kwyxz: We can definitely create a specific case for Solaris

kwyxz: Wait isn't it already what we're doing here? 

`ifneq ($(findstring SunOS,$(shell uname -a)),)`

Am I looking at the wrong commit?

fr500: My apologies if I missed it, I'll check in the morning

fr500:
I tested RetroArchievements and it seems they are working fine.
I still have to refactor multicart, I don't like the current implementation very much

fr500:
Updated a few things and yep removed that memmaps implementation it's not required for achievements, we may tackle it again later

libretro/Makefile
fr500: @kwyxz thing is, the SHARED flags were changed for solaris but that is being used for all unixes apparently

fr500:
```Makefile
ifneq (,$(findstring unix,$(platform)))
   TARGET := $(TARGET_NAME)_libretro.so
   fpic := -fPIC
   ifneq ($(findstring SunOS,$(shell uname -a)),)
   CC = gcc
   SHARED := -shared -Wl,--version-script=link.T -Wl,-z,defs
   else
   SHARED := -shared -Wl,--version-script=link.T
   endif
   ifneq ($(findstring Haiku,$(shell uname -a)),)
      LIBS :=
   endif
```

I think this would be enough?

fr500: or is it the other way around? not good on makefile fu

kwyxz: I might be wrong but here's how I understand the current Makefile: the ifneq line is a double negative
````
   ifneq ($(findstring SunOS,$(shell uname -a)),)
   CC = gcc
   SHARED := -shared -z defs
   else
   SHARED := -shared -Wl,--version-script=link.T -Wl,-z,defs
   endif
````
If the findstring command looking for "SunOS" does not return "empty" it means we're building for Solaris hence `CC` becomes `gcc` and SHARED becomes `-shared -z defs` (gcc on Solaris does not handle commas for some reason)
Otherwise SHARED is `-shared -Wl,--version-script=link.T -Wl,-z,defs`

We could replace `uname -a` with `uname -s` though, to avoid hostnames that would include `SunOS` even though people doing this are looking for trouble IMHO

PR #352, opened 2018-07-01
Libretro changes **DO NOT MERGE**

fr500:
**Do not merge**, just want to highlight some of the changes to see if they are upstreamable, I'm willing to clean them up and backport them if at all possible.

The travis stuff and makefile changes were shouldn't be backported at all.

I'm basically talking about @mudlord's interpolation changes @Dwedit's changes and @Tatsuya79's changes here.


Dwedit:
Regarding SPC_DSP.cpp/.h:

Hard Audio Disable was a bad idea, and I should probably take it out of the Libretro version altogether.  While it can provide a speed boost in certain circumstances, it does it by shutting off emulation of the DSP entirely.  It's only meant to be used in situations where audio will never be used again without loading state first, and even then it can probably cause problems.  I might consider restoring it if I can find a valid use-case where its performance boost means the difference between fullspeed and not.

EnqueueSamples and GetOutputBufferBase were leftovers from when I was trying to fix issues with savestates and the sound buffer.  After making a solution that stored some of the sound buffer into the savestate, I later realized I could just call the audio_callback function instead, and not need any of that stuff.  It's dead code that should be removed from the Libretro version.

As for the Savestate code...

What was meant by a "Fast" savestate?  It means that loading state would be done in-place rather than using separate copies of the memory buffers.  When I was profiling the savestate code, it turned out that doing the memory copies for the temporary copies of the memory buffers ate up significant time, so loading states in-place was the solution to that.

However, if you load states in-place, you lose the ability to proceed with the old game state if loading state fails.  You could get some error-resistance back with basic validation (nothing unexpected, incomplete, or missing) of all the chunks before loading the state in-place.

Additionally, if buffers are loaded in-place, you need to skip the code that clears the memory after it loads them.

The other thing "Fast" savestates did was skip most of the Reset code.  PPU reset went down to just a few tasks (reinitialize dirty tiles) to prevent screen garbage.

Saving and Loading the framebuffer (local screenshot) was causing problems in the Libretro version.  Since the framebuffer was not saved, the code that tried to load the framebuffer would never have any data to load, so it would blank out the screen.  This had some unwanted side effects.  If you were paused, and loaded state, you always ended up with a black screen afterwards.

So I had changed the code for the Libretro version to skip blanking out the screen.

Maybe the options related to saving and loading the frame buffer could be made into settings?  The three possible new settings are: Option to Save framebuffer, Option to Load Framebuffer if present in savestate, and Option to Clear Framebuffer if absent from savestate.

I also wrote some code for the savestate system to skip past some unexpected chunks.

OV2:
What's the use-case for the fast snapshots? Is there a platform where the current code is too slow?

Saving the screen is already an option (Settings.SnapshotScreenshots). If it's not present it will blank out the screen, yes - but I don't see how that is any worse than keeping the old image from before the state. Shouldn't be a problem to switch that to a tri-state setting if skipping that is of some benefit.

We can maybe take the regular overclocking in, since it looks like it just requires defines in port.h and some ifdeffing in snes9x.h

I'll defer to @bearoso regarding the interpolation, I'm not that versed in audio details.

Dwedit:
The use case of fast savestates is for the RunAhead feature of RetroArch, which aims to reduce input lag by displaying a frame from the future instead of the current frame.  It saves and loads state 60 times per second in the worst case.  So anything that can speed up savestates is a welcome feature for that.

bearoso:
Interpolation is a no. The gaussian interpolation is part of the actual hardware's output. We're not going to cater to people craving the nostalgia of ZSNES's shitty sound.

The overclocking, if it goes in, definitely needs to be an ifdef. Turning those constants into variables for heavily accessed values is a speed hit.

fr500:
Just to be clear, I'm not pushing these changes upstream, just up for consideration, so maybe a cleaner better implementation may be done in upstream. If/when that happens I'll remove these commits.

This guy backported the LLE emulation for Cx4 and DSP1b I was considering incorporating that too in the branch.
https://github.com/kps501/bright_snes9x/blob/master/libretro/libretro.cpp#L212

w4st3d:
@bearoso 
1) I must say I am.....perplexed you feel that way. To be honest I am **not allowed** to say what I truly feel and think why this decision was done. 

 2) The **completely optional** audio interpolation was done at the DSP level so that then it doesn't need to be done by a outside DSP plugin if you turn off the interpolation entirely. Sinc and cubic interpolation has been tested to have **tangible audio quality** benefits over Gaussian interpolation and I don't see why there is such a reaction to it when Snes9x is not aiming for per-cycle accuracy to be honest. 

3) Nearest and linear can definitely be ignored, but the **instant hardline no answer** when it comes to any audio DSP changes seems weird to me for audio quality reasons.

qwertymodo:
I agree that allowing interpolation to be optional, possibly with other interpolation options like sinc and cubic, is a nice idea. I mean, we have half a dozen video filters, why not allow something similar on the audio? Definitely keep Gaussian enabled by default though.

Tatsuya79:
I find it interesting to hear how it renders without any interpolation too.
Even if you don't play like this.

bearoso:
It definitely [i]needs[/i] interpolation. A no interpolation option is out. I guess I'm not completely against adding the other interpolation options if they indeed sound much better, but personally I don't see the need. Let's just try and not complicate things by adding a ton of different ones. No more than one or two extra options.

bearoso:
The cubic interpolation option is out. We can't use ZSNES code in Snes9x like that since it's GPL and Snes9x isn't. The linear option is wrong; the samples aren't properly biased with respect to the offset.

DAOWAce:
@bearoso > Interpolation is a no. The gaussian interpolation is part of the actual hardware's output. We're not going to cater to people craving the nostalgia of ZSNES's shitty sound.

What about Snes9x's sound prior to version 1.52?  Because this is the exact thing I experienced when I decided to look into my emulating days of old and saw Snes9x was being updated: The sound now sounds like muffled shit and there's no way to change it.

This is practically the only thread I can find about it: http://www.snes9x.com/phpbb3/viewtopic.php?f=6&t=6922

2013 it was complained about.  I'd have expected a change in the last 4½ years..

_Also, nobody can register on the forum now because the recaptcha is deprecated._

1.51 and ZSNES offer higher quality sounding audio due to this.  Sure, it may not be accurate, but it SOUNDS better, and there's no reason to not give people a choice.. a choice we've had for over a decade.

Emulators can do something **better** than the original console and not be a problem, y'know.  Just look at ePSXe, PCSX2 and Dolphin.  Widescreen support on non-widescreen games, anti-aliasing, anisotropic filtering, massively increased internal resolution, 60FPS support on 30FPS games.. so many improvements that make playing on the original console an objectively worse experience _(not to mention load times, save states and unlimited save slots)_.  I would **never** pull my original consoles out of storage to play them over an emulator.


That aside, as much as I'd like to stick to 1.51 (or zsnes), the sound emulation isn't on the same level as the modern Snes9x.  It speeds up and slows down at random and can crackle at times.. plus DirectDraw and Windows 10's issues..

I'd really love to come back and experience my SNES days, but seeing all these problems now, I doubt I ever will.

bearoso:
>The sound now sounds like muffled shit and there's no way to change it.
>1.51 and ZSNES offer higher quality sounding audio due to this. Sure, it may not be accurate, but it SOUNDS better, and there's no reason to not give people a choice.. a choice we've had for over a decade.

That muffled sound is how the SNES hardware actually sounds. Interpolation is necessary to get that effect. In some games, disabling interpolation might make things seem a little sharper, but it also reveals glitches that were covered up by the original hardware. (Listen to the FF6 WoR overworld music in ZSNES. Bleh :-P).

Hopefully we'll have those extra options brought to parity with the libretro core soon so we're one codebase. We'll have the interpolation options available then somehow, even if just hidden in the config file.

> This is practically the only thread I can find about it: http://www.snes9x.com/phpbb3/viewtopic.php?f=6&t=6922 Also, nobody can register on the forum now because the recaptcha is deprecated.

We have no control over the forums. That's why development is mostly centered here now.

OV2:
The forum registration should work again, I've notified ryan and he removed the reCaptcha.

DAOWAce:
> Hopefully we'll have those extra options brought to parity with the libretro core soon so we're one codebase. We'll have the interpolation options available then somehow, even if just hidden in the config file.

Thank you, that sounds great. Even a .conf only variable would be wonderful. Hope it happens soon! :)

I tried importing stuff into VS2017 but with the way github works with forks and whatnot (doesn't fork PR's) it all just seemed way too complicated to even understand for me.  Even building failed and I couldn't figure out how to link zlib/libpng.. Gave up and succumbed to using RetroArch _(and boy is it not made for KB/M.. or even ease of use IMO)_. Maybe one day I'll learn this stuff and not feel like an idiot. :(

Thanks for the forum fix as well.

bearoso:
Ok, so what else do we need to do here?
I know of: 
* Overclocking options
* Sprite limit
* Fast save states
* Pull request #373 maybe

I fixed the Chrono Trigger thing where the screen jumps in the core, so we don't need to mess with that. For the Lufia 2 hack I'd just get rid of it.

The "hard audio disable" option is not going in. Simply stopping the DSP like that is a bad idea. I'm surprised it works--and the reason is I suspect some aspect of the runahead code makes it moot.

OV2:
The blargg filter seems a bit weird in this in-between layer, but I guess I don't mind. What does the retroarch team think?

I can take over the fast savestates - unless you're already working on that bearoso ?

fr500:
I don't really see a point for blargg filters since libretro has both CPU filters and GPU shaders but it's up to you.

Other than that, that would be all, we would be fully up-to-date, I'd transfer the repo to @twinaphex and hope it doesn't deviate like the other one :)

fr500:
@Tatsuya79 pointed out this hack that is required for deae tonosama to work
https://github.com/fr500/snes9x/issues/1

bearoso:
> I can take over the fast savestates - unless you're already working on that bearoso ?

Nope, you can do it if you want.

>I don't really see a point for blargg filters since libretro has both CPU filters and GPU shaders but it's up to you.

Now that you mention it, that does seem kind of pointless, unless the person isn't using RetroArch as the libretro frontend.

>Edit: Oh, I'm missing to MACs Rifle support.

I was going to let someone else do that one. Is the multi-rom stuff up-to-date?

> @Tatsuya79 pointed out this hack that is required for deae tonosama to work
fr500#1

Wasn't that just broken by all the extra mapping stuff? We removed most of it.

fr500:
Multi-rom is working fine yes (for sufami at least).
Not 100% clear on satellaview and friends because I don't know how to test.

`Wasn't that just broken by all the extra mapping stuff? We removed most of it.`
@Tatsuya79 can you try the latest master and see if it works?

Tatsuya79:
Just got the copy protection screen (compiling master from here).

turhope:
Tried RA CPU blargg filters but it strangely chops game window in 1/2 (KDL3 hires areas). GPU NTSC is nice but has different shades and texture compared to blargg (and leans overly heavy when stacking more shaders).

Okie dokie. I'll close the PR since both parties feel it's redundant.

fr500:
@turhope well actually @bearoso has a point, not all libretro frontends have the filters, and we have at least two other cores that have them so no harm I think

bearoso:
> Just got the copy protection screen (compiling master from here).

Ok, added the change. 

> The sprite limit hack isn't working as it's named incorrectly.

Whoops. Fixed.

AxlRocks:
Don't know if it's alright to ask question/sort of make a request/suggestion here, I figure it's better now than after it merges, but could the overclocking option be modified to allow users to tweak each value independently? I don't know if each value is tied together in more intricate ways than they appear, I only seen the values changing over the course of various builds and presumed they could be quite tweakable, if not having been built for libretro options standards.

The reason I ask is because it would be nice to be able to "fix" problem games by maximizing the values that can be overclocked yet leaving alone the ones that cause issues, instead of having libretro styled, non-informative, "user friendly", consolized, predefined options that, as mentioned, may cause issues but could be fixable otherwise. 

Though rare, I have seen some situations like that with the near 300 tests I've done of NES overclocking in Mesen/FCEUX, though to be fair, that's much more of a miracle solution that has incredible compatibility, so I was hoping with this being a fair bit less compatible with the default options, perhaps more fine tuning per-game could solve some issues.

Or maybe I'm just totally wrong and only very specific values will work together or whatever. If so, nevermind, but if at all possible I think it's best in a standalone emulator to have more open-ended options wherever possible.

fr500:
it's "merged already" as in libretro fork was reset to upstream.

AxlRocks:
Just seen that, nevermind then. Deleting post.

PR #354, opened 2018-07-07
fix crash with sufami turbo and remove lingering runahead code

fr500:
This fixes a crash with Sufami turbo when SRAM autosave is enabled in RetroArch (basically data was returning null).

Also removes some old runahead code I didn't notice.

---

Saving for slot B still seems wrong, we create on file per-slot so you can load different combinations, so I created two copies of SD Ultra Battle - Ultraman Densetsu (sd1.sfc, sd2.sfc)

It creates the SRAM "properly" for slot a, but when I reset and load again slot b is still empty.
It can be worked around by loading in reverse order (sd2 in slot a, sd1 in slot b).

If you guys can give me a hand with how should I get the data and size properly I could fix this.

### DIFF
--- a/libretro/libretro.cpp
+++ b/libretro/libretro.cpp
@@ -151,7 +151,7 @@ void retro_set_environment(retro_environment_t cb)
 
     static const struct retro_subsystem_rom_info multicart_roms[] = {
         { "Cart A", "smc|sfc|swc|fig|bs", false, false, false, multi_a_memory, 1 },
-        { "Add-On B", "smc|sfc|swc|fig|bs", false, false, false, multi_b_memory, 1 },
+        { "Cart B", "smc|sfc|swc|fig|bs", false, false, false, multi_b_memory, 1 },
     };
 
     static const struct retro_subsystem_info subsystems[] = {
@@ -1241,6 +1241,7 @@ void* retro_get_memory_data(unsigned type)
 
     switch(type) 
     {
+        case RETRO_MEMORY_SNES_SUFAMI_TURBO_A_RAM:
         case RETRO_MEMORY_SAVE_RAM:
             data = Memory.SRAM;
             break;
--- a/libretro/libretro.cpp
+++ b/libretro/libretro.cpp
@@ -1107,17 +1107,6 @@ static void report_buttons()
 {
     int offset = snes_devices[0] == RETRO_DEVICE_JOYPAD_MULTITAP ? 4 : 1;
 
-    int runahead_poll = 0;
-    int result = -1;
-    if(environ_cb(RETRO_ENVIRONMENT_GET_AUDIO_VIDEO_ENABLE, &result))
-    {
-            // runahead off: 3
-            // runahead secondary off: 0,0,0,0,0,3
-            runahead_poll = (result & 0x0b)==3;
-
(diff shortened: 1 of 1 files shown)


kps501:
Mostly kept my suggestions dry upstream because I'm always unsure about what's correct (ex. achievements). I have something similar to this in my private build. But because I cheated by combining all sram together into st_ram tag together, which is probably bad idea.


      case RETRO_MEMORY_SNES_SUFAMI_TURBO_A_RAM:
         data = Multi.sramA;
         break;
      case RETRO_MEMORY_SNES_SUFAMI_TURBO_B_RAM:
         data = Multi.sramB;
         break;


      case RETRO_MEMORY_SNES_SUFAMI_TURBO_A_RAM:
         size = Multi.sramSizeA ? (1 << (Multi.sramSizeA + 3)) * 128 : 0;
				 break;
      case RETRO_MEMORY_SNES_SUFAMI_TURBO_B_RAM:
         size = Multi.sramSizeB ? (1 << (Multi.sramSizeB + 3)) * 128 : 0;
				 break;

fr500:
That's the first thing I tried but it didn't work.
I still get the B slot message

kps501:
I can't translate the Japanese error message entirely but it mentions how Slot B cartridge is not formatted and you must run it separately first to initialize the game pak memory. Then the multi-link function will work.

So I think you already correctly knew how to fix the mainline libretro port. ;)


That's how bsnes+, snes9x behaves also. Game A. Then Game B (+ optional Game A). But I've never seen actual real hardware video, so I wouldn't close case just yet. ^^

OV2:
Saving sramB looks correct to me. How are you loading sufami turbo via retroarch?

kps501:
Via Retroarch, you place stbios.bin in /system folder. Then load st game1. When it boots, you go to menu and select "Load Multi-Cart Link" (this sounds stupid and I should've called it "Load Multi-Cart" -- upvote renaming like fr500 mentions). Select st game1. Then select st game2. Then select run again. It then loads slotted cartridge A/B.

OV2:
I was still using rmenu, that's why I did not see any option. But even when using "Load Multi-Cart Link" I only get calls to retro_load_game, never to retro_load_game_special (tried latest stable and nightly). The "--subsystem" cli option also doesn't seem to work, ra.exe simply exits immediately.

fr500:
Hey @OV2 there are still a few bugs with it, can't load from zips using the "load" option, just via the browse option, if you use browse it loads the game right away, maybe that was the case.

The GUI subsystem implementation is... well new.

From cli it should be retroarch -L core rom1 --subsystem multicart_addon rom2

OV2:
Ok, it worked via cli. The help says "--subsystem=", which is what I tried. The position also seems to influence the slot order for some reason? "-L core rom1 --subsystem multicart_addon rom2" loads rom2 to slot A, while "--subsystem multicart_addon -L core rom1 rom2" loads rom1 to slot A.

It looks to me like it works as intended. @fr500 the SRAM for the cart in slot B never gets initialized by the sufami turbo, you need to place it in slot A for that.

fr500:
ahh cool then we're golden there! thanks for letting me know..

I'll try to whip some sense into the cli options sometime.

PR #356, opened 2018-07-08
fix mouse and lightguns

fr500:
another fix for mouse and lightguns, some code was missing from that runahead poll attempt


OV2:
Sorry for the late merge, I somehow forgot about it...

fr500:
np, thank  you.

PR #368, opened 2018-07-28
gtk: Use datarootdir instead of datadir.

orbea:
Currently snes9x ignores `--datarootdir` and only uses `--datadir`. This is a problem because I would like to change the install path for `cheats.bml` without changing the path for the icons and desktop file.

So using `--datarootdir` sets the path for all data including locales, icons, the desktop file and `cheats.bml` while `-datadir` only sets the path for `cheats.bml` which can override `--datarootdir` for that one file.

This seems to work as desired here, but please review to make sure its right.

Edit: From `./configure ---help`.
```
  --datarootdir=DIR       read-only arch.-independent data root [PREFIX/share]
  --datadir=DIR           read-only architecture-independent data [DATAROOTDIR]
```

### DIFF
--- a/gtk/data/Makefile.am
+++ b/gtk/data/Makefile.am
@@ -1,42 +1,41 @@
-desktopdir = $(datadir)/applications
+desktopdir = $(datarootdir)/applications
 desktop_DATA = snes9x.desktop
 
-icon16x16dir = $(datadir)/icons/hicolor/16x16/apps
+icon16x16dir = $(datarootdir)/icons/hicolor/16x16/apps
 icon16x16_DATA = snes9x_16x16.png
 
-icon24x24dir = $(datadir)/icons/hicolor/24x24/apps
+icon24x24dir = $(datarootdir)/icons/hicolor/24x24/apps
 icon24x24_DATA = snes9x_24x24.png
 
-icon32x32dir = $(datadir)/icons/hicolor/32x32/apps
+icon32x32dir = $(datarootdir)/icons/hicolor/32x32/apps
 icon32x32_DATA = snes9x_32x32.png
 
-icon64x64dir = $(datadir)/icons/hicolor/64x64/apps
+icon64x64dir = $(datarootdir)/icons/hicolor/64x64/apps
 icon64x64_DATA = snes9x_64x64.png
 
-icon128x128dir = $(datadir)/icons/hicolor/128x128/apps
+icon128x128dir = $(datarootdir)/icons/hicolor/128x128/apps
 icon128x128_DATA = snes9x_128x128.png
 
-icon256x256dir = $(datadir)/icons/hicolor/256x256/apps
+icon256x256dir = $(datarootdir)/icons/hicolor/256x256/apps
 icon256x256_DATA = snes9x_256x256.png
 
-#iconscalabledir = $(datadir)/icons/hicolor/scalable/apps
+#iconscalabledir = $(datarootdir)/icons/hicolor/scalable/apps
 #iconscalable_DATA = snes9x.svg
 
 cheatsdir = $(datadir)/snes9x
 cheats_DATA = ../../data/cheats.bml
 
 install-data-hook:
-	mv -f $(DESTDIR)$(datadir)/icons/hicolor/16x16/apps/snes9x_16x16.png \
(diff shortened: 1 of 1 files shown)


orbea:
Any comments?

bearoso:
> I would like to change the install path for cheats.bml without changing the path for the icons and desktop file.

Why? Where do you plan on putting it?

orbea:
My idea would be to use `--datadir=/usr/share/games`, the result as shown in my built package.
```
usr/share/
usr/share/applications/
usr/share/applications/snes9x.desktop
usr/share/games/
usr/share/games/snes9x/
usr/share/games/snes9x/cheats.bml
usr/share/icons/
usr/share/icons/hicolor/
usr/share/icons/hicolor/128x128/
usr/share/icons/hicolor/128x128/apps/
usr/share/icons/hicolor/128x128/apps/snes9x.png
usr/share/icons/hicolor/16x16/
usr/share/icons/hicolor/16x16/apps/
usr/share/icons/hicolor/16x16/apps/snes9x.png
usr/share/icons/hicolor/24x24/
usr/share/icons/hicolor/24x24/apps/
usr/share/icons/hicolor/24x24/apps/snes9x.png
usr/share/icons/hicolor/256x256/
usr/share/icons/hicolor/256x256/apps/
usr/share/icons/hicolor/256x256/apps/snes9x.png
usr/share/icons/hicolor/32x32/
usr/share/icons/hicolor/32x32/apps/
usr/share/icons/hicolor/32x32/apps/snes9x.png
usr/share/icons/hicolor/64x64/
usr/share/icons/hicolor/64x64/apps/
usr/share/icons/hicolor/64x64/apps/snes9x.png
usr/share/locale/
usr/share/locale/es/
usr/share/locale/es/LC_MESSAGES/
usr/share/locale/es/LC_MESSAGES/snes9x-gtk.mo
usr/share/locale/fr_FR/
usr/share/locale/fr_FR/LC_MESSAGES/
usr/share/locale/fr_FR/LC_MESSAGES/snes9x-gtk.mo
usr/share/locale/ja/
usr/share/locale/ja/LC_MESSAGES/
usr/share/locale/ja/LC_MESSAGES/snes9x-gtk.mo
usr/share/locale/pt_BR/
usr/share/locale/pt_BR/LC_MESSAGES/
usr/share/locale/pt_BR/LC_MESSAGES/snes9x-gtk.mo
usr/share/locale/ru/
usr/share/locale/ru/LC_MESSAGES/
usr/share/locale/ru/LC_MESSAGES/snes9x-gtk.mo
usr/share/locale/sr@latin/
usr/share/locale/sr@latin/LC_MESSAGES/
usr/share/locale/sr@latin/LC_MESSAGES/snes9x-gtk.mo
```
It would not be correct to install `share/locale`, `share/icons` and `share/applications` to `share/games`, but some other distros may have more specific needs (GuixSD and NixOS maybe?).

bearoso:
Ok, I understand that.

orbea:
@bearoso Meson broke this, it seems to be a problem inherent with meson which doesn't know the difference between datarootdir and datadir.... :(

bearoso:
Should be easy enough to fix. Let me work out the bugs.

bearoso:
Ok, I added an option 'appdatadir' that can specify where the stuff that only Snes9x uses will go. (cheats.bml for now)
It should work if you specify -Dappdatadir=/usr/share/games/snes9x on the meson setup/configure line. Otherwise it uses the standard. Meson uses FDO-compliant paths, and datadir matches that spec, so swapping that meaning to match autotools would be confusing.

orbea:
Thanks, it still seems like a meson bug to not distinguish between project specific files and more general data files, but this will work...

bearoso:
Generally, that stuff goes in /usr/share/*appname*, so you're asking for a special case. I know some distributions like Gentoo do a special games directory, but FDO frowns upon that. Personally, I think if the user wants something different, then it's OK. I do think the notations appdatadir and datadir are more specific than datadir and datarootdir. And meson is worlds easier to understand and use than autotools.

orbea:
I wouldn't argue for autotools, personally I would of chosen sofort. For example as used in slibtool.

https://git.midipix.org/cgit.cgi/slibtool/tree/sofort

The main benefit for `--datarootdir` and `--datadir` is that they were standard for all projects, instead of like cmake where every project reinvents the wheel. The specific names don't really matter that much...

PR #371, opened 2018-07-30
ppu.windowleft fix

turhope:
some games incorrectly show rogue vertical line on column 0


bearoso:
Thanks. I knew about some glitches with column 0 but hadn’t looked into it.

PR #373, opened 2018-08-06
libretro: blargg filters

turhope:


turhope:
Saw this in snes9x bright (libretro), tried it, and backported here.

Issue #382, opened 2018-08-08
Request: sprite limit


ramtrooper:
Reducing sprite flicker option is very nice to have.

Can we also have an option to display more than 32 sprites per line? Some racing, shooting games have tons of particle effects or lots of trees,hazards that can annoyingly popup or disappear.

Thank you again.

OV2:
This is already in the libretro version. We can probably add a config option to activate it in other ports.

ramtrooper:
Tried the libretro version and it's nice! Would be great to have in other ports.

Although .. I think there's 2 types of sprite limiter?
- 34 tiles per line (lots of big sprites) (* = libretro only)
- 32 sprites per line (lots of 8x8 sprites) (# = n/a?)

Tatsuya79:
I'm the one who did that change for libretro.
It's done a bit poorly from 34 to 60 = what looked the most ok-ish in Gradius 3 [without touching gfx.h as I feared that would change the default behaviour when the option is off.](https://github.com/Tatsuya79/snes9x/commit/2387e04463bd37752b1264b947f78ba4808c485e)
I wasn't sure how to do that properly, please revisit it if you know better!

jjfmz42015:
If sprite limit removal is done please make it optional. This will break some games.

bearoso:
Of course it's optional. 

> 32 sprites per line (lots of 8x8 sprites) (# = n/a?)

I don't think increasing this will affect much. If you can find a case where increasing the tile max doesn't work, we'll allow increasing the actual object limit as well.

> It's done a bit poorly from 34 to 60 = what looked the most ok-ish in Gradius 3 without touching gfx.h as I feared that would change the default behaviour when the option is off.
I wasn't sure how to do that properly, please revisit it if you know better!

It's perfectly fine. And I don't think those gfx.h changes would have any side-effects.

jjfmz42015:
Different platform but taitos chase hq on pce comes to mind they used the sprite limit to hide things. I imagine other programmers on other platforms may do similar things.

bearoso:
My comment was a little indirect. We're _not_ going to increase the sprite limit by default.

ramtrooper:
The hakchi snes9x core uses a hacked dsp-4 bios where this upper limit works -- Top Gear 3000, stage 1, # tiny visible trees on horizon.

I thought Phalanx has a flicker which isn't solved by tiles per line. And R-Type. I'll play lots of games and see if I can come up with a concrete list in a few months.

[...]

Issue #380, opened 2018-08-08
Request: cleaner hires blending


ramtrooper:
I saw someone playing snes mini with Kirby's Dream Land 3 (hakchi snes9x). But screenshots show the hires blending is somehow different from normal or libretro snes9x -- it's a lot less blurry without filters, plus works correctly with shaders like scale2x.

I would love to have this clean hires blending -- I heard there's some hack that gives the same non-blurry mixing look, but it doesn't work the same when used with scale2x.

Do you know how they did this?

Thank you

Tatsuya79:
[Probably something from here.](https://github.com/kps501/bright_snes9x/blob/master/libretro/render.cpp#L272)

bearoso:
> Probably something from here.

That was copied from the GTK+ port. I just merge adjacent fields in pairs instead of a spatial blur to the right like bsnes and the Windows port.

Tatsuya79:
Oh? I didn't know.
Something I noticed when snes9x is displaying a higher res (in Hoshi no Kirby 3), the left side of the screen has like a 1 pixel column with "artefacts".

bearoso:
>Oh? I didn't know.
Something I noticed when snes9x is displaying a higher res (in Hoshi no Kirby 3), the left side of the screen has like a 1 pixel column with "artefacts".

This is probably due to the ppu window fix. I wonder if byuu's hires plotter will work now and fix that.

bearoso:
Kirby 3 should be fixed. Depending on the BG offset, column 0 wasn't being updated most of the time, so garbage was left on the screen. @OV2, you might want to look and see if the change I made is OK.

Tatsuya79:
Thanks that looks better now!
[Does it mean this buffering can be removed now](https://github.com/libretro/snes9x/commit/0c5cee96eaa2446b3bd40adf36a61d02e1d6e35b) as it was made to try to hide that issue (somewhat)?

I removed it locally and I didn't see any difference in kirby 3 hi-res clouds scrolling.

bearoso:
> Does it mean this buffering can be removed now as it was made to try to hide that issue (somewhat)?

I guess. I don't understand why that's there.

bearoso:
Ok, I added the merge option, and got rid of the buffers and merged your sprite limit increase to 128.

Tatsuya79:
Thanks, works fine!

ramtrooper:
[...]

PR #386, opened 2018-08-11
rise sprite limit hack to 128, remove hi-res buffer (libretro)

Tatsuya79:
Following https://github.com/snes9xgit/snes9x/issues/382 I was testing the sprite limit hack again and realised a change in gfx.h wasn't needed at all.

Rising the current setting to 128 is enough to get rid of most flickering in Gradius 3.


Tatsuya79:
...and remove "hi-res blending buffer" that shouldn't be needed now.
(see https://github.com/snes9xgit/snes9x/issues/380)

PR #393, opened 2018-08-15
hires blending

retrotalker:
automatic game option: Hoshi no Kirby 3, Kirby's Dream Land 3, Jurassic Park (all)


OV2:
What's the purpose of the automatic switch? Usually you either use hires blending, or you use a filter that does it by itself.

retrotalker:
Originally thinking about people who don't know when to use "merge" blending filter, or maybe "merge-auto" name fits better. There's so few hires games that benefit from the merge blending and look awful when used wrongly (or forget to turn off).

The automatic switch is just a helper, but I guess it'd turn snes9x more into pcsx2 like hacks database which would be terrible idea just for this special case.

Yeah, I guess you can close this if the benefit looks small.

retrotalker:
Was politely told this is a bad idea.

PR #399, opened 2018-08-18
Fix several file references in the APU folder

bladeoner:


bladeoner:
I found another reference in smp_state.cpp so it took a little bit longer to finish the pull request.

After the Appveyor build check and if the devs agree this changes can be merged.

Thanks in advance.

PR #400, opened 2018-08-18
interpolation - clamp output

retrotalker:


bearoso:
Are the other interpolation methods actually overflowing? Because part of the reasoning for including them was that they naturally wouldn't.

retrotalker:
(using libretro)

The cubic and sinc more -- sometimes you can hear random small (soft) pops during the Zelda 3 rainstorm (indoor, outdoor) and other games had random noise static. I threw in a debug volume meter and saw overflows.

Also tried using a weighted hermite at 0.25 intervals and suffered basically the same thing.

Gaussian can overflow (seen it) but that's already covered.

Linear I don't know if it overflows; I imagine it wouldn't but eh.

Doing a clamp resolved everything for me and figured it was safe to do. *shrug*

PR #402, opened 2018-08-18
activate controllers

retrotalker:
Looney Tunes B-Ball multitap


retrotalker:
libretro doesn't attach controllers until after S9xControlsReset is done, which creates some detection errors. So we force snes9x to connect the new device immediately.

PR #404, opened 2018-08-19
interlace load state

retrotalker:


retrotalker:
Regarding A.S.P. bug report with interlace and load states, didn't see anything wrong with win32. libretro has 1 frame of old interlace data but much less severe than earlier.

OV2:
The duplicate code should probably move to a new function. Also, wouldn't it require to save GFX.InterlaceFrame / Timings.InterlaceField in the snapshot?

retrotalker:
snapshot.cpp
- static FreezeData SnapTimings[] =
  INT_ENTRY(6, InterlaceField)
- FreezeStruct(stream, "TIM", &Timings, SnapTimings, COUNT(SnapTimings));
- UnfreezeStructCopy(stream, "TIM", &local_timing_data, SnapTimings, COUNT(SnapTimings), version);

so I think Timings.InterlaceField is covered by snapshots.

Yeah, it does make sense to make a new function for dupe gfx code.


Question:
I'm wondering if it'd be more useful to use 213F for checking interlace field instead everywhere?

I was always quietly assuming that GFX.InterlaceFrame = Timings.InterlaceField = Memory.FillRAM[0x213F] & 0x80? Or is it possible that they all can be different and not tied to each other?

retrotalker:
Moved duplicate code to new function: S9xGraphicsScreenResize. Tested win32 build and no interlace loadstate problems anywhere so far. Timings is part of snapshot so that's covered.

Unsure what to edit next. If this looks okay-like, I'll make a new pr with 1 squashed commit.

Issue #378, opened 2018-08-08
Request: change lightgun cursor


ramtrooper:
I saw some versions of snes9x allow changing lightgun icon, even using custom ones. Also I like being able to use the 50% transparent custom colors as it's easier to see through.

Thanks

OV2:
You can change the lightgun type and colors in the config file:

```
[Controls]
MacsRifleCrosshair = 4 Red/Orange
```
Where it is TYPE FOREGROUND/BACKGROUND

Valid names are
```
Mouse1Crosshair
Mouse2Crosshair
SuperscopeCrosshair
Justifier1Crosshair
Justifier2Crosshair
MacsRifleCrosshair
```

Crosshairs go from 0 to 16
Custom ones are also possible.

Colors are
```
        "Trans",
	"Black",
	"25Grey",
	"50Grey",
	"75Grey",
	"White",
	"Red",
	"Orange",
	"Yellow",
	"Green",
	"Cyan",
	"Sky",
	"Blue",
	"Violet",
	"MagicPink",
	"Purple",
	NULL,
	"tBlack",
	"t25Grey",
	"t50Grey",
	"t75Grey",
	"tWhite",
	"tRed",
	"tOrange",
	"tYellow",
	"tGreen",
	"tCyan",
	"tSky",
	"tBlue",
	"tViolet",
	"tMagicPink",
	"tPurple"
```

ramtrooper:
Oh, that's nice to use!

Does libretro port also have a config file?

ramtrooper:
Thanks for making this work in libretro! But gui says 25% Gray, 50% Gray, 75% Gray -- these colors don't work.

I think it's a typo?
- 25% Grey, 50% Grey, 75% Grey
- 25% Grey (blend), 50% Grey (blend), 75% Grey (blend)

Otherwise this can be closed soon as solved!

ramtrooper:
Works perfectly now! Thanks guys!

ramtrooper:
I think it'd help others if those configuration instructions are written into the win32 readme or auto-generated into the conf file somehow.

I'm migrating off libretro to soak in what the standalone experience is like, and I don't want to forget how to do this later on.

Thanks for all the team keeps on doing for us! :)

PR #406, opened 2018-08-21
lightgun cursor grey typo

retrotalker:


OV2:
Isn't this just AE / BE ? I don't mind either way, maybe a native speaker can give a preference.

bearoso:
Yeah, this is personal preference. Gray is usually used in the US, grey is used in the UK, and they're both correct.

retrotalker:
controls.cpp uses 25Grey,50Grey,75Grey so I guess we could stick with that? There's also "The Grey (2011)" movie.

retrotalker:
Forgot to mention that this commit is for
https://github.com/snes9xgit/snes9x/issues/378

libretro GUI says Gray but strcmp(Grey) so 25,50,75 don't work atm.

PR #419, opened 2018-08-28
fix the rest of the include file references

bladeoner:
Checked and corrected all files if the #include references.


bearoso:
Just glancing at this, I see an error where you broke compilation. Is there a reason why you can't just use include directories?

bladeoner:
Hi @bearoso, there is no reason that I can't use include directories.

I saw in the APU folder some references which needed to be corrected and while I was busy with that I thought to check all the files. 

The main reason to do this is that I wanted to contribute to the snes9x project.

bearoso:
@bladeoner: filter_epx_unsafe.cpp

@OV2, do you think this is a good idea?

OV2:
I don't really mind one way or the other. It might be more readable to have them all be relative, but then you always need to adjust them in a file if you move it. And you still need to have the current dir on the include path.

PR #420, opened 2018-08-28
Change detection of copier headers

Dwedit:
The current header detection system breaks when nonheadered scores are in specific value ranges. 
Nonheadered scores should not matter at all when considering the presence or absence of a header.

I also added in two more metrics for detecting headers:  File size of exactly a multiple of 64K + 512 bytes, and first 512 bytes containing up to 32 nonzero bytes.

edit: Scores for those metrics are completely arbitrary, and have not been methodically tested against synthetic headered roms designed to fail both tests, or nonheadered roms with exactly 512 bytes of padding added onto the end.  Also should I use `bool8` instead of `bool` inside of functions?

### DIFF
--- a/memmap.cpp
+++ b/memmap.cpp
@@ -1380,6 +1380,20 @@ int CMemory::ScoreLoROM (bool8 skip_header, int32 romoff)
 	return (score);
 }
 
+int CMemory::First512BytesCountZeroes() const
+{
+	const uint8 *buf = ROM;
+	int zeroCount = 0;
+	for (int i = 0; i < 512; i++)
+	{
+		if (buf[i] == 0)
+		{
+			zeroCount++;
+		}
+	}
+	return zeroCount;
+}
+
 uint32 CMemory::HeaderRemove (uint32 size, uint8 *buf)
 {
 	uint32	calc_size = (size / 0x2000) * 0x2000;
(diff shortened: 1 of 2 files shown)


bearoso:
bool should be fine as long as we don't care about its size. 

I'm wondering about appended firmware. The closest in size I see is the gameboy boot rom, which is 256 bytes, so one wouldn't be a problem.

OV2:
Usually headers are removed when loading from file in HeaderRemove. Maybe it would help adding this to the libretro loading? I see there's even a custom function in libretro.cpp that does the same, but is not used for regular roms.

Dwedit:
Didn't notice the HeaderRemove function, as the libretro version never called it.  It seems that HeaderRemove just assumes if the file size is a multiple of 8K + 512, there is always a header to remove.  So if you added 512 bytes of junk onto a headerless rom, it would assume that was a header.

Dwedit:
Maybe I should add a commit that replaces HeaderRemove entirely with this?

OV2:
I ran a test with ~4000 unheadered roms, where it only triggered for 16, mostly ST or Beta Roms. There were no cases where the old variant worked and the new failed, which was my main concern. Unfortunately I don't have a large headered set to test against, so I can't tell how those are affected.

Dwedit:
If you don't have a large headered collection to test, there are still some other possible tests to do:

* Append 512 bytes of zeroes to the beginning (basically a copier header)
* Append 512 bytes of nonzero bytes to the beginning
* Append 512 bytes to the end
* Append 512 bytes to both the beginning and end

This will help test the effect of the size heuristic (multiple of 64K + 512) and zero byte population heuristic.

PR #423, opened 2018-08-29
MacOSX: explicitly set default values for new options

tmkk:
MacOSX port should explicitly set default values for new configuration options added in fdae8cc, 6d15bf7 and cf5681a.


OV2:
Thanks for keeping an eye on this.

Issue #418, opened 2018-08-27
Daffy Duck - The Marvin Missions, screen flickering


Tatsuya79:
Screen flickers randomly during gameplay.

Current git version, RA win7 x64 Nvidia.

bearoso:
We had this one earlier. It was IRQ-related, but I don't remember which issue it was.

bearoso:
This is interesting. IRQs and NMIs are occurring where they're supposed to during those black frames. Forced blank is turned on/off appropriately and so is brightness.

telviseo:
irq 208 = turn screen off, black bar bottom
irq 011
irq 014 = turn screen on


Here's when it blinks on us:
*** WRAM Refresh  HC:0544
$80:F206 8D 00 42    STA $4200  [$81:4200]   A:0031 X:0036 Y:0032 D:0000 DB:81 S:01E8 P:envMXdizc HC:0584 VC:014 FC:04  HV .n.    HT:602 VT:14 C:602
--- IRQ Timer HC:602 VC:14 set 357970 cycles HTimer:1 Pos:0148->0602  VTimer:1 Pos:014->014
--- IRQ Timer Enable HTimer:1 Pos:0602  VTimer:1 Pos:014
$80:F209 68          PLA                     A:0031 X:0036 Y:0032 D:0000 DB:81 S:01E8 P:envMXdizc HC:0608 VC:014 FC:04  HV ...    HT:602 VT:14 C:357970
$80:F20A C2 20       REP #$20                A:0012 X:0036 Y:0032 D:0000 DB:81 S:01E9 P:envMXdizc HC:0634 VC:014 FC:04  HV ...    HT:602 VT:14 C:357970


1 frame of black, missed irq


*** WRAM Refresh  HC:0548
$80:820B AD 8E 1E    LDA $1E8E  [$00:1E8E]   A:0000 X:0000 Y:0002 D:0000 DB:00 S:01F4 P:envmXdiZc HC:0594 VC:014 FC:05  HV .n.    HT:602 VT:14 C:602
Timer triggered

--- IRQ Timer HC:628 VC:14 set 357966 cycles HTimer:1 Pos:0148->0602  VTimer:1 Pos:014->014
*** IRQ
[...]

PR #431, opened 2018-09-08
daffy duck irq non-reschedule

lacastri:
fix random flickering black screens


lacastri:
Fixes this game but I'm not 100% sure this is hw-accurate.

bearoso:
I don't think that would break anything. I'll have to test to find out.

bearoso:
Seems to work fine. Would probably be more hw-accurate as well.

Dwedit:
Just linking this to issue #418

PR #446, opened 2018-11-05
Remove memory leak in loadzip.cpp

bonimy:
* Call `unzClose(file)` outside of the assert clause. In Release mode,
any statement in an assert clause is not called, causing a memory leak.
If the file does not close successfully, `assert(FAIL)` is called.


bearoso:
Cherry-picked this without all your merges.

bonimy:
Pointing to 6591ee9 to reference this PR

PR #456, opened 2018-12-01
libretro: Allow overriding -flto.

orbea:
With some users it could be handy to set `-flto` manually, for example with gcc its possible to get a speed up for linking with `-flto=jobserver`.

I added the `+` since gcc documentation claims its needed, its part of the POSIX spec so it should not hurt?
```
     You can also specify '-flto=jobserver' to use GNU make's job server
     mode to determine the number of parallel jobs.  This is useful when
     the Makefile calling GCC is already executing in parallel.  You
     must prepend a '+' to the command recipe in the parent Makefile for
     this to work.  This option likely only works if 'MAKE' is GNU make.
```
From the `gcc.info` page.

The `make(1)` posix spec.
```
+
    If the command prefix contains a <plus-sign>, this indicates a makefile command line
    that shall be executed even if -n, -q, or -t is specified.
```
http://pubs.opengroup.org/onlinepubs/9699919799/utilities/make.html

Note that `-flto=` options are not compatible between gcc and clang so that we can't force `-flto=jobserver` in the Makefile unfortunately.

The credit for the idea goes to the current snes9x.SlackBuild maintainer (Also includes the libretro core), I just implemented the changes.

### DIFF
--- a/libretro/Makefile
+++ b/libretro/Makefile
@@ -2,6 +2,7 @@ DEBUG = 0
 HAVE_EXCEPTIONS = 0
 HAVE_STRINGS_H = 1
 
+LTO ?= -flto
 SPACE :=
 SPACE := $(SPACE) $(SPACE)
 BACKSLASH :=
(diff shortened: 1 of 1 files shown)


orbea:
Note that if anyone objected to `-flto` before with this they could now remove it using `make LTO=`.

PR #491, opened 2019-02-07
APU Refactor

bearoso:
This is a big cleanup of the sound output code. All drivers have been tested and are working properly, aside from 8-bit, Monaural, and Reverse Stereo, which I removed from apu.cpp.
Do we need any of those?
* 8-bit I think we can discard completely. If there's ever a port that needs it, it can downsample on its own. 
* Mono sound I think is also unnecessary. Most, if not all, current single-speaker devices do sample combining automatically.
* Reverse stereo is used on the Wii. 

These are postprocesses. Everything in the core has been 16-bit stereo since 1.52, so nothing uses them until S9xMixSamples. I'd like to keep S9xMixSamples unambiguous, meaning it always outputs one sample, 16-bit, left then right, alternating. 

@OV2, do you want to keep any of those three options in the Windows port, or is it OK to remove them?


OV2:
It's fine to remove them, I don't think anyone actually uses any of those options. If a port really needs reverse stereo then they can probably do it themselves.
One thing I noticed on a quick check: if you only enable sound sync, with no vsync and no automatic framerate, then it runs at ~90fps.
DSound now sounds broken on my pc, I'll try to look into that this weekend.

bearoso:
> DSound now sounds broken on my pc, I'll try to look into that this weekend.
I thought it was fine, but I'll check it. I'm still trying some things out. I'm going to see if I can get the DSP to just write straight to the resampler buffer without all the shuffling.

bearoso:
What I've done now is make the DSP and MSU1 emit samples directly to their resampler buffers. This cuts the number of buffers from 5 (DSP, DSP Resampler, MSU1, MSU1 resampler, Platform API) to 2. 

Ideally, the resampler buffer is as small as can be and the output API should try to keep the buffer empty. With sound sync on GTK, I only check that once per frame, which means that it needs to guarantee 535 sample frames are empty for the next one, so that's the current minimum. I'm thinking about making it a bit more granular, but that entails moving the sound sync code into the drivers. NBD I guess, since DynamicRateControl is there.

DirectSound through the timer callback isn't going to work. The resampler is NOT thread safe, but giving access to the Snes9x core means we can't lock it. What's currently in the DirectSound code is legacy from when Snes9x just generated samples on-demand, so I'm thinking of just removing the timer callback and doing the same thing during the Snes9x samples_available callback.

SDL's audio callback API won't work either, and their push API works like crap, so I might need to just remove it.

OV2:
The problem with DirectSound (and why I never removed the timer) is that it doesn't simply stop playing if it has consumed the currently available samples, but loops its buffer continuously. So when emulation is paused it needs something that fills the buffer with silence, otherwise you get a really choppy repeat of the last samples. It also unfortunately has really inaccurate read/write cursors, that is why I made the system of blocks where it simply tries to fill the previous block once the read cursor appears in the next block.

bearoso:
If needed, we can have the direct sound driver keep it’s own separate big buffer that it can lock, and just push all the available data to that when samplesavailable is called.

OV2:
I'm also not completely opposed to remove dsound entirely if it proves too much of a hassle. It's a nice fallback since it is always available, but I'm not sure that many people actually use it.

bearoso:
winmm might be a good replacement fallback. It’s pretty much the same core as dsound these days, but a bit simpler.

bearoso:
GTK's drivers all working now. I think if we do a proxy buffer, like I did with SDL, for DirectSound, it should work OK.

I'm still getting a little bit of scratchyness on Windows/XAudio2 that I can't pinpoint. With the frequency at 48000Hz everything seems to be mostly OK, but moving down to 44100Hz some weirdness appears. It gets worse the larger the buffer size goes. 32ms is mostly clean, but 64 has intermittent static, and 128+ it's pretty frequent. This was there before the refactor, too, and I was hoping it'd clear up, but no such luck.

OV2:
Weird, I can't hear any issues even with 210ms and 32khz.
Maybe we need to increase the number of chunks that can be passed to xaudio when the buffer size increases?

bearoso:
32000 seems fine, too. I was thinking it was something to do with 44100 being a weird multiple.

OV2:
I tried it at work and also didn't notice any particular scratchiness - maybe it's just my hearing...
Only with low buffer sizes and dynamic input rate enabled, but that's to be expected I guess.

Should I move the sound sync into the drivers? And the goal is to allow an even smaller resampler buffer?

bearoso:
I think it’s just temperamental Windows drivers. Moving the check into the driver just makes it less likely to overflow or underflow, and it’s a pretty small change for Xaudio.

I’m experimenting with dsound. I’m going to try and duplicate byuu’s method and see if it works ok.

bearoso:
DirectSound is a pain in the ass. No wonder it was replaced. I wrote a WaveOut driver and it's already proving better for me. I propose we use it to replace DirectSound. It's only been a thin layer over Core Audio since Vista, and built-in compatibility runs back to 98.

OV2:
Sure, I have no problem with that. I think I introduced xaudio2 back then exactly because the old dsound didn't work correctly for me and I had trouble fixing it up.

I've removed the dsound files, hooked up WaveOut and moved the sound sync into the drivers. We could probably also change the drivers to use variable size blocks - do you think that would be worth it?

bearoso:
> We could probably also change the drivers to use variable size blocks - do you think that would be worth it?

I thought about it, having a max block size, and after that just increasing the number of blocks to get up to the total buffer size. But at the moment things are working with all drivers, so I'd rather not mess it up. I think a fixed 8 will be fine, but if you want to play around with it, it's OK.

bearoso:
Since everything's working again, I'm going to go ahead and merge this.

qwertymodo:
Is there any reason we're still using 32040 as the APU base sampling frequency?  That number is literally just an average of a very small sample size of consoles taken by a couple of people on their consoles.  The ceramic resonator has a nominal frequency of 2.576MHz, which is exactly 32,000 * 768.  It has always seemed odd that we used that particular rate rather than the nominal one.  We use nominal values for everything else (main CPU frequency, pixel RGB values, etc), and if we wanted to represent component tolerances, we'd really need a much larger sample size to pick a better default value.

bearoso:
It would entail changing the SPC and DSP frequency and it might break something, but we could try it. I'd have to recalculate the ratio fractions for the CPU/SPC.

bearoso:
Ok, I changed it. If something breaks, we'll revert.

PR #497, opened 2019-02-16
Fix SPC_DSP GCC compile error

bladeoner:


bearoso:
Where are you needing this? SPC_DSP.cpp isn't meant to be compiled on its own. We fixed the unix port.

bladeoner:
On the port Snes9x EX+ (Android) I get compilation errors with file SPC_DSP.cpp:

In file included from ./src/snes9x/apu/bapu/dsp/SPC_DSP.cpp:5:
./src/snes9x/apu/bapu/dsp/SPC_DSP.h:25:19: error: unknown type name 'Resampler'
        void set_output( Resampler* resampler );
                         ^
./src/snes9x/apu/bapu/dsp/SPC_DSP.h:140:2: error: unknown type name 'Resampler'
        Resampler *resampler;

If it can be fixed in a alternative way I would be happy.

For the second change the SPC_DSP.h this is not causing compilation errors but I think it needs to be reviewed, the code used in SPC_DSP.cpp is different:
SPC_DSP.cpp: void SPC_DSP::set_output( Resampler *resampler )
SPC_DSP.h  (original): void set_output( Resampler* resampler );
SPC_DSP.h (the change I made): void set_output( Resampler *resampler );

bearoso:
> If it can be fixed in a alternative way I would be happy.

I already told you: don't compile SPC_DSP.cpp on its own. It's included by sdsp.cpp.

Where the asterisk goes doesn't matter.

bladeoner:
@bearoso your suggestion worked, thank you.

Also good to know where the asterisk goes doesn't matter.

PR #505, opened 2019-02-23
APU refactoring follow-ups for Mac port

tmkk:
Mac port works best when `Settings.SoundInputRate` is set to 31950. 31920 tends to buffer overrun and 32000 tends to underrun. I'm not sure why.

Commit b2fe66e made change on OS-independent file, so please check if it doesn't break compilation for other platforms.

### DIFF
--- a/apu/resampler.h
+++ b/apu/resampler.h
@@ -9,7 +9,11 @@
 
 #include <cstring>
 #include <cassert>
+#if __cplusplus >= 201103L
 #include <cstdint>
+#else
+#include <stdint.h>
+#endif
 #include <cmath>
 
 class Resampler
--- a/macosx/mac-audio.mm
+++ b/macosx/mac-audio.mm
@@ -26,6 +26,7 @@
 #include <AudioToolbox/AudioToolbox.h>
 #include <AudioUnit/AudioUnitCarbonView.h>
 #include <pthread.h>
+#include <semaphore.h>
 
 #include "mac-prefix.h"
 #include "mac-dialog.h"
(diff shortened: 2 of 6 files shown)


OV2:
Doesn't sound sync prevent the overrun? Or was this just in your test for a good default value?
Did you try the dynamic rate control option? I saw you added the parts in the sound code, but I'm guessing there is currently no way to enable it in the mac port?

__cplusplus has a really old value in msvc by default, but I fixed that on the msvc side.

Thanks again for keeping this up to date!

tmkk:
Sound sync helps but it causes occasional frame dropping (as expected). DRC prevents over(under)run on a host side buffer, but it generates clicks constantly (causing overrun in snes9x side buffer, I guess). That's why I haven't added the interface to enable DRC yet.

bearoso:
The problem is that it depends on your display. Apple tends to get the refresh rate near 60 with first-party displays, which means 32000 * 60 / 60.09881 ~= 31947. 

Third party displays are a crapshoot. 1920x1080 or 1920x1200 displays are most common and tend to be around the NTSC rate 59.94 or 59.95, which is why I chose the 31920 number. I also implemented an option to check the refresh rate and choose the correct number on X11 and Windows so the default should go unused most of the time with those.

bearoso:
Looking at the Mac audio code, it's kind of screwy. The callback is sending a list of buffers to render to, but Snes9x is only filling one. Effectively, it's using a tiny audio buffer. The "interval" option is used for setting fragment size independent of the buffer size, which is probably increasing contention and bottlenecking the streaming process. The whole thing obviously hasn't changed since the old "generate sound on-demand" method of Snes9x 1.51 and earlier.

tmkk:
From what I understand, AudioBufferList is structured as a list for non-interleaved channel format (LLLLLL... + RRRRRR...). When interleaved the number of buffers to be filled is always one.

bearoso:
That makes sense. Apple’s documentation is as bad as microsoft’s. There’s not much specification,  just examples to copy. 

I’d get rid of the “mixing interval,” though, and make it something proportional, like 1/8 of the buffer size. I think accumulating big portions of the buffer is swinging the dynamic rate control too far.

tmkk:
Actually we can't control the interval accurately. When the interval is set to 16ms (kAudioUnitProperty_MaximumFramesPerSlice = 512), a render callback requires 371-372 (not constant) frames per call on my machine. I suppose only suboptimal performance can be achieved under "push buffers from an emulator core to system" strategy with CoreAudio.

Issue #461, opened 2018-12-16
Slowdown snes9x 1.57 libretro


marleyap:
Hello users snes9x Retroarch , wanted to do a little information about a drop of frames in snes9x 1.57 in the game Megaman X2, more precisely in the DEEP SEA BASE stage, in the part where I left that big robot fish, this fall of frames happens in TV box amilogic s905x 1.5 GHz, this drop is in the range of 2 fps less than full speed, now I do not know if it is due to lack of hardware of my tv box or lack of core optimization snes9x in tv box s905x, I wonder if programmers could solve this problem , even because the FX games run smoothly, without dropping frames ... Thank you all!

marleyap:
I'm having the core updates snes9x 1.57 in the download options nightly Linux ARMHF, I use the Lakka OS 2.1.1 version for TV android box with AMILOGIC S905X processors, 1.51 GHZ AND with 1 GB RAM, I tested on two TV box models, the two present slowndonw in the certain bubble crab phase of the game Megaman X2, in the part when it appears that when Big robot fish the game has fall in up to 58 fps, and also tested with old standard SNES9X core that comes in the LAKKA OS in the case of Snes9x 1.54 .1 frame crashes are worse, reaching up to 54 fps, now my doubt is if this happens for lack of optimization of the updated snes9x core for TV box s905x or lack of TV box hardware that I use. Stranger than games with Fx chips that are much heavier to emulate than Megaman X2's Cx4 chips and without the use of the video threading option run like a feather without gagging anything at all.

marleyap:
(update) I just tested the update snes9x libretro, which is already version 1.58, slowndonw has decreased slightly, now it is in 1.2 fps less than full speed, I try to drop up to 58.8 fps, still give a feeling slight noise and slowness in the audio in the game and part of the game I quoted earlier in my TV box. I hope they correct it soon I think it is a small lack of optimization of the core for the processor of my device ...

bearoso:
I don't know why that part would be CPU-intensive. The optimization level could probably be bumped up to -O3 without any problems, though.

orbea:
@marleyap Could you possibly bisect this issue and see if any specific commits introduced this issue?

marleyap:
[...]

PR #523, opened 2019-04-02
Unmacro tile.cpp

yoffy:
This patch provides maintainability and debuggability of tile.cpp.
It replacements many macros to C++ template, and abolish self inclusion.


bearoso:
I was thinking about doing this before you posted it. 

I fixed the compiler error and tested it and you're getting a 3-4% total speed loss. With Snes9x, we're out of simple things to optimize, so that's a significant amount. You need to figure out what's doing that before we can commit it. I looked through the assembly, and the templating is working correctly. GetCachedTile is a vtable lookup so that might be slow.

yoffy:
> you're getting a 3-4% total speed loss.

What OS and compiler did you test?
And, how can I measure it? (framerate, CPU cycles, benchmark test, etc.)

My environment is Raspberry Pi 3B, armv7 1.2GHz x 4core, Linux 4.19 (Raspbian 9.8), gcc 6.3.0.
It doesn't look like it's getting slower in this environment.
I also tried compiling with Visual C ++, but I got an error that 'zconf.h':No such file or directory.

bearoso:
Linux x86_64, clang and gcc 9. LTO and non-LTO. It's slightly slower (max framerate and CPU cycles). You'd need to measure it because it's not visibly obvious. The simplest way is to just unthrottle everything on something static and see how many FPS you get.

OV2:
On win32 x64 with my somewhat imprecise measurement (movie playback with no syncs, averaging fps) I see about 6% speed loss. SMW2 intro and turning island (PAL version), 481fps vs 452fps.

yoffy:
We can include "snes.hpp" and define `alwaysinline` every method. Maybe it generates same code as macro implementation.
However, the size of tile.o implemented by the macro is up to 2 MB. I think too large code will cause frequent cache misses.
And I think it's better to carefully tune for slower CPUs like Raspberry Pi, than for CPUs that are already performing well.
Anyway, it may be better to use alwaysinline for this pull request and separate to tune for another pull request.

yoffy:
I'm defined alwaysinline to functions.

bearoso:
I'm still seeing a 1.8% speed loss with alwaysinline.

> And I think it's better to carefully tune for slower CPUs like Raspberry Pi, than for CPUs that are already performing well.

I agree. We barely managed to get a 6% increase in performance in issue #461 to help lower-end CPUs, and I'd rather not take it away so soon.

I really like this idea because it makes the code much cleaner and compiles faster, but we can't sacrifice too much performance.

This is a big change, so we won't merge until after 1.60, so there's no hurry. Get some numbers on your Raspberry Pi to see the difference. Try the Yoshi's Island 2 rotating island with an uncapped framerate like OV2. Or try the Rockman X2 areas mentioned in issue #461.

bearoso:
It seems Clang likes these changes, and it yields a small **increase** in performance. That's pretty good. I'll have to look at the MSVC numbers, but I think performance is good enough to merge in 1.61.

OV2:
We could also leave the macroed version in as an alternative for ports. Unless we really overhaul the whole tile file at some point it shouldn't be that much of a maintenance issue to keep them up to date.

yoffy:
Raspberry Pi 3B and gcc 6.3.0 was also slightly slower. (40 fps-> 39 fps)
I think that caused by the compiler loading both B and C while trying to remove branches such as `A ? B : C`.
But I could not fix it.

yoffy:
The compilation of tile on unix has been made much faster. In Raspberry Pi 3B (gcc 6.3.0) compilation of tile.cpp equivalent will be 7.8 times faster.

master branch (51edec7):
```
$ time make ../tile.o
:
real    6m17.435s
user    6m15.691s
sys     0m1.091s
```

This patch (7603ed0):
```
# parallel compilation
$ time make -j3 ../tile.o ../tileimpl-n1x1.o ../tileimpl-h2x1.o ../tileimpl-n2x1.o
:
real    0m48.442s
user    2m6.856s
sys     0m1.708s

# sequential compilation
$ time make -j1 ../tile.o ../tileimpl-n1x1.o ../tileimpl-h2x1.o ../tileimpl-n2x1.o
:
real    1m48.678s
user    1m47.046s
sys     0m1.305s
```

bladeoner:
@ov2 @bearoso if you want I can put this in the port Snes9x GX and share it with the community to get some quick feedback. If it works fine for that port it's not necessary to maintain the macroed version for it.

I say this because the resources in the Wii are really limited. For your information it's running core 1.60 with the latest Snes9x fixes but the 1.52/1.53 APU to balance out the used resources.

Thanks in advance for your reply.

bearoso:
Thanks. It'll be helpful to see the effect on slower hardware. I think we're probably going to change to this wholesale and try to make up any difference after.

bladeoner:
Here are my test results for compiling Snes9x GX, without the changes:
real    23m5.593s
user    0m46.338s
sys     1m59.481s

With the changes:
real    17m55.635s
user    0m45.903s
sys     2m2.520s

So it seems we reduce some compilation time with that also.

I also tried few games myself and I barely notice a difference, so that is positive but lets wait for some more feedback on GBA Temp.

bladeoner:
@bearoso gpt some test results back, 10 games where tested and where running more or less the same as before this change. Only one game Kirby's Dreamland 3 has some huge framedrops but that's an known issue and has nothing to do with the change.

So I would say merge the commit as it has a lot of benefits, like compilation speed but also reduction in file size of the emulator. From 5,5 MB to 5 MB for Snes9x GX.

PR #539, opened 2019-05-15
libretro uwp

funbars:


funbars:
note: uwp broken in master.

```cl : Command line error D8048 : cannot compile C file '../filter/snes_ntsc.c' with the /ZW option```

PR #542, opened 2019-05-27
libretro android makefile (fix)

funbars:


hizzlekizzle:
I tested this and it corrects the color format mismatch with blargg's filter.

Issue #533, opened 2019-05-02
Winter Gold (PAL), Player character glitches on start


Pat86:
When you start a race, your chracter glitches a bit right after the intro sequence. Same happens sometimes after you crash and the little "video sequence" plays.

https://streamable.com/4s9xe

bearoso:
This is a problem with the SuperFX bump we did for Yoshi's Island. Decrease the SuperFX clock speed a little bit in settings.

Pat86:
Aren't both games using the same Super FX 2 Chip? Ergo same clock speed. How come it isincreased it for Yoshis Island?

bearoso:
Yes, that's why Winter Gold changed, because we increased the SFX2 speed. Yoshi's Island was running too slowly. We've got 3 SFX2 games, and I didn't think the slight increase would affect the other two. Clearly, I was wrong.

bladeoner:
@Pat86 could you please try this one with Winter Gold (E) [snes9x-1.60-24-g880d9fa-win32-x64.zip](https://github.com/snes9xgit/snes9x/files/3234576/snes9x-1.60-24-g880d9fa-win32-x64.zip) and see if this works for you?

@bearoso could you please try the above file with Yoshi's Island?

I made a change (a hack) that will always use the old values before the SuperFX bump except for Yoshi's Island. If it works I can open a PR.

The changes can be found here https://github.com/bladeoner/snes9x/commit/880d9fac69bdf7b56afba7c164aa754c8d8da895

Thanks in advance.

Pat86:
@bladeoner 

Just tried it with my own dump, works flawlessly. no more glitched character model at the beginning of the race.

bladeoner:
@Pat86 thanks for the quick response.

bladeoner:
@Pat86, bearoso reverted the change which should solve the Winter Gold issue:
Revert "SFX: Make SFX2 1/3 faster than 1.59.2."

themalu866:
The characters still glitch when the race starts.
But only in these two.
I marked the two where the glitch happens.
![uiy7uty](https://user-images.githubusercontent.com/42208976/69049532-e2c25e00-09de-11ea-9e97-c8e449599f61.png)
![rt](https://user-images.githubusercontent.com/42208976/69049593-05ed0d80-09df-11ea-9e62-bad33b7a5daf.png)
[...]

PR #543, opened 2019-05-31
Fix SuperFX2 Core Speed

bladeoner:
This will fix https://github.com/snes9xgit/snes9x/issues/533


bearoso:
Not going to add a game hack for this. It barely affects Yoshi's Island, so I'm just reverting the speed completely.

PR #552, opened 2019-07-16
Fix GTK build on PowerPC systems with Altivec

shawnanastasio:
On the PowerPC family of architectures with Altivec enabled
(ppc, ppc64, ppc64le), one of snes9x's dependencies will
include the altivec.h header which redefines bool and vector.
Since this interferes with the C++ data types of the same
name, they must be undefined to prevent the build from failing.

### DIFF
--- a/gtk/meson.build
+++ b/gtk/meson.build
@@ -6,7 +6,7 @@ project('snes9x-gtk',
 args = ['-DSNES9X_GTK', '-DUNZIP_SUPPORT', '-DNETPLAY_SUPPORT', '-DJMA_SUPPORT', '-Wall', '-W', '-Wno-unused-parameter']
 srcs = []
 deps = []
-includes = ['../apu/bapu', '../', 'src']
+includes = ['../apu/bapu', '../', 'src', '../ppc']
 warns = []
 
 prefix = get_option('prefix')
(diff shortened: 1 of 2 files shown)


bearoso:
I’d like to know what dependency is causing this. Try putting a #error in your system altivec.h and get the stack of includes that loads it.

shawnanastasio:
It seems like SDL is the culprit:
```
                 from /usr/include/SDL2/SDL_cpuinfo.h:55,
                 from /usr/include/SDL2/SDL.h:38,
                 from ../src/gtk_control.h:13,
                 from ../src/gtk_config.h:14,
                 from ../src/gtk_s9x.h:10,
                 from ../src/gtk_sound_driver.h:10,
                 from ../src/gtk_sound_driver_alsa.h:11,
                 from ../src/gtk_sound_driver_alsa.cpp:7:
```

For what it's worth, I've had to implement similar workarounds in other projects like [supertux](https://github.com/SuperTux/supertux/pull/1056).

shawnanastasio:
It seems like [this was fixed upstream in SDL](https://hg.libsdl.org/SDL/rev/6da95eef84a2), but the latest stable release 2.0.9 doesn't have the fix. It's up to the snes9x team whether they want to include this workaround to allow compilation against stable and older SDL releases on PPC.

bearoso:
Ok, this should be fixed by undefining those after each SDL.h include. include_next is a GCC extension, so I'd definitely like to avoid that.

PR #599, opened 2019-11-18
CI

kivutar:


OV2:
Some details would be nice. I'm guessing you want automated travis builds for all libretro platforms in this and your branch?
But you cannot simply kill the appveyor builds, we still want those.

kivutar:
Sorry, this was not been to be merged, just a test to see if I can build on different platforms in PRs.

PR #600, opened 2019-11-18
libretro vita building

tryal-star:


tryal-star:
That r-value reference is what's biting the Vita gcc toolchain. The psp toolchain is worse and doesn't have C++0x support (?) for r-value reference.

OV2:
Not sure where you see an rvalue here, XPSdecode is always called with a variable as parameter.

tryal-star:
It's likely the arm-vita-gcc compiler. You can see the log here:
https://ci.appveyor.com/project/tryal-star/snes9x/builds/29117226#L213

The "fix" is kinda stupid imo but it makes it work for Vita. But I just read your Windows Appveyor logs and notice the other error. Ouch. That's a close pr.


I added a PSP target to the makefile and I know that it didn't like the pass-by-reference either.
```static uint32 XPSdecode (const uint8 *data, unsigned &addr, unsigned size)```

But I don't know of another way around this gcc platform compiler bug?

PR #602, opened 2019-11-23
Mac merge

MichaelBuckley:
Here's a minimal Mac port as discussed in #586. It's not great code, but it works for now. I hope to continue development on it in 2020.


srtcemu.cpp
```diff
@@ -3,6 +3,8 @@
  * Copyright (c) byuu
  *****/
 
+#include "snes9x.h"
+#include "srtcemu.h"
```
OV2: Any particular reason for this? srtcemu.cpp is included in srtc.cpp and shouldn't be compiled by itself.

MichaelBuckley: That is a good catch. Nothing outside of the `macosx` directory should have been altered, and I don't recall making this change. It's possible Xcode decided to be "helpful" here. I will revert and make sure everything still compiles.

MichaelBuckley: Sure enough, that change did not need to be there. I'm removing it now.

OV2:
Thanks a lot for your work! I have no real way of doing a test compile, but I'll merge it anyway.
Do you want to provide a binary build that users can already test?

MichaelBuckley:
Thank you for merging! I haven't made any changes since the build I posted in the Github issue. Users can use that build for testing. https://github.com/snes9xgit/snes9x/files/3834437/Snes9x.app.zip

bearoso:
Michael, since you’re obviously committed, I’ve invited you as a collaborator with write access to the repository.

MichaelBuckley:
@bearoso Thanks! I will probably still send pull requests for any code that lands outside of the macosx directory, just to make sure it has another set of eyes on it, but I don't expect there will be much.

PR #621, opened 2020-01-13
Mac: make HID hat switches work

gtalusan:
I have an old 8BitDo NES30 Pro gamepad.  It's detected by the master build of Snes9x, but the button preferences dialog doesn't pick up the D-Pad.

After some digging, I see that the D-Pad is a kHIDUsage_GD_Hatswitch as opposed to a bunch of buttons as they are on my 8BitDo SN30 Pro gamepad.

The existing code sets up a mask for the state of the hat switch, but I believe the HID descriptor enumerates each of the values.  i.e., the value of each hat switch will range from kIOHIDElementMinKey to kIOHIDElementMaxKey.

0 - UP
2 - RIGHT
4 - DOWN
6 - LEFT

with intermediate values indicating both buttons are pressed, i.e., 1 is UP+RIGHT and 5 is DOWN+LEFT.

I've changed the code to support an 8 state hat switch based on this behaviour.

@MichaelBuckley - I'm not sure of the origin of the bit mask.  Please let me know if this breaks any of your controllers.

### DIFF
--- a/macosx/mac-joypad.mm
+++ b/macosx/mac-joypad.mm
@@ -34,31 +34,6 @@
 #include "mac-os.h"
 #include "mac-joypad.h"
 
-#define	kUp(i)					(i * 4)
-#define	kDn(i)					(i * 4 + 1)
-#define	kLf(i)					(i * 4 + 2)
-#define	kRt(i)					(i * 4 + 3)
-
-#define	kPadElemTypeNone		0
-#define	kPadElemTypeHat4		1
-#define	kPadElemTypeHat8		2
-#define	kPadElemTypeAxis		3
-#define	kPadElemTypeButton		4
-#define	kPadElemTypeOtherHat4	5
-#define	kPadElemTypeOtherHat8	6
-
-#define	kPadXAxis				1
-#define	kPadYAxis				0
-#define	kPadHat					0
-
-#define	kMaskUp					0x0800
-#define	kMaskDn					0x0400
-#define	kMaskLf					0x0200
-#define	kMaskRt					0x0100
-
-typedef	hu_device_t		*pRecDevice;
-typedef	hu_element_t	*pRecElement;
-
 std::unordered_set<JoypadDevice> allDevices;
 std::unordered_map<JoypadDevice, std::map<uint8, std::map<int8, S9xButtonCode>>> defaultAxes;
 std::unordered_map<JoypadDevice, std::map<uint8, S9xButtonCode>> defaultButtons;
@@ -175,7 +150,7 @@ void gamepadAction(void *inContext, IOReturn inResult, void *inSender, IOHIDValu
         objcInput.value =inputStruct.value;
 
         pthread_mutex_unlock(&keyLock);
-        if (info.min != info.max)
+        if (info.usage != kHIDUsage_GD_Hatswitch && info.min != info.max)
         {
             if (inputStruct.value <= info.min || inputStruct.value >= info.max)
             {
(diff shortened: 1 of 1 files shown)


macosx/mac-joypad.mm
```diff
@@ -177,7 +152,7 @@ void gamepadAction(void *inContext, IOReturn inResult, void *inSender, IOHIDValu
         pthread_mutex_unlock(&keyLock);
         if (info.min != info.max)
         {
-            if (inputStruct.value <= info.min || inputStruct.value >= info.max)
+            if (inputStruct.value >= info.min && inputStruct.value <= info.max)
```
gtalusan: Ugh, actually I see why this code was like this now... will need a bit of work/thinking to fix the breakage.

MichaelBuckley:
@gtalusan That bitmask comes from the older joypad code. It's not something I introduced.

I do know from other ptojects I've worked on that _most_ hats follow the scheme where up is 1, right is 2, down is 4, and left is 8, with the diagonal directions being OR'd between them. (so e.g. up-right is 3, down-left is 12), but not all of them work that way.

The old code appeared to assume all 8-way joypads worked that way, but had special exceptions hard-coded into the soruce for a couple of joypads. My goal with the rewrite was to make it customizable so it would work with any hat, but I was unaware of any hat that uses 0 to represent a direction. What value does it send when releasing the hat, so that no direction is pressed?

gtalusan:
@MichaelBuckley the released/centered state on my controller is 15, so it tripped over the bit mask code and all buttons were pressed.

I hunted around GitHub for some example HAT switch code on MacOS and stumbled across this just now in RetroArch https://github.com/libretro/RetroArch/pull/4426 and libSDL https://github.com/dhewg/libsdl/blob/master/src/joystick/darwin/SDL_sysjoystick.c#L787.

They're both slightly similar in spirit to my pull request here.

MichaelBuckley:
Interesting. I will test this out tonight. We probably can't merge this as-is without breaking some hats, but I'll look to the code you pointed out for inspiration.

gtalusan:
I think this and the previous code were probably already broken for 4-way hats.  I searched through the git history and found this code:

```
+                                                                                               {
+                                                                                                       if (((command % 4 == 0) && (curv == 0)) ||      // Up    : 0
+                                                                                                               ((command % 4 == 1) && (curv == 4)) ||  // Down  : 4
+                                                                                                               ((command % 4 == 2) && (curv == 6)) ||  // Left  : 6
+                                                                                                               ((command % 4 == 3) && (curv == 2)))    // Right : 2
+                                                                                                               gDirectionInfo[i].type = kPadElemTypeOtherHat8;
+                                                                                                       else
+                                                                                                               gDirectionInfo[i].type = kPadElemTypeHat8;
+                                                                                               }
+                                                                                               else
+                                                                                               {
+                                                                                                       if (((command % 4 == 0) && (curv == 0)) ||      // Up    : 0
+                                                                                                               ((command % 4 == 1) && (curv == 2)) ||  // Down  : 2
+                                                                                                               ((command % 4 == 2) && (curv == 3)) ||  // Left  : 3
+                                                                                                               ((command % 4 == 3) && (curv == 1)))    // Right : 1
+                                                                                                               gDirectionInfo[i].type = kPadElemTypeOtherHat4;
+                                                                                                       else
+                                                                                                               gDirectionInfo[i].type = kPadElemTypeHat4;
+                                                                                               }
+                                                                                       }

```

The enumeration for an 8-way matches the code in this pull request.  I think a 4-way hat can be supported by this patch by multiplying the reported value by 2 if the reported range is 0-3.

gtalusan:
Here's another example from Dolphin: https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/InputCommon/ControllerInterface/OSX/OSXJoystick.mm#L293

MichaelBuckley:
@gtalusan I borrowed a joypad from a friend which I've used for testing in the past, one which I was sure reported up as 1, but it worked with this code. Unless you're aware of any additional issues we need to consider, I think it's safe to merge this. If we later hear from someone who it doesn't work for, we can investigate at that point.

gtalusan:
Thanks @MichaelBuckley.  I'm good with helping support the code and fixing any issues that come with it.

PR #632, opened 2020-02-18
[appveyor] Fix "git: No tags can describe xxx" error caused in forked repositories

gocha:
The `git describe --tags` command in appveyor.yml will cause a "git : fatal: No tags can describe" error in a forked repository if there are no appropriate tags. That command is just used to build the filename of the artifact.

This PR removes the command from appveyor.yml and uses some [predefined environment variables](https://www.appveyor.com/docs/environment-variables/) instead.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -43,8 +43,7 @@ build_script:
   - if "%config%"=="libretro Release" msbuild libretro\libretro-win32.vcxproj /t:build /p:Configuration="%config%";Platform="%platform%" /m /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
 
 after_build:
-  - ps: $env:gitrev = git describe --tags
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -43,8 +43,7 @@ build_script:
   - if "%config%"=="libretro Release" msbuild libretro\libretro-win32.vcxproj /t:build /p:Configuration="%config%";Platform="%platform%" /m /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
 
 after_build:
-  - ps: $env:gitrev = git describe --tags
-  - ps: $env:my_version = "$env:gitrev"
```
bonimy, line 47: What did `$env:gitrev` show and why is it should it be changed to `version-commit` format?

OV2:
Sorry for the long wait - never got around to take a closer look.

PR #644, opened 2020-06-21
Improvements to mute when using turbo

### LINKED ISSUE #643: Pausing and unpausing during turbo ignores mute setting
If I have `Mute sound when using turbo` enabled, and I pause then unpause while turbo is enabled, the sound will become unmuted.

I will take this opportunity to point out that this setting should also apply to rewind, as it isn't nice to listen to music in reverse, in the same way it isn't nice to listen to it sped up.

Edit: This happened on snes9x 1.60, GTK+ version 87

AlexApps99:
It will now also mute during rewinding, and after pausing.

This will fix #643.

### DIFF
--- a/gtk/po/es.po
+++ b/gtk/po/es.po
@@ -1296,12 +1296,14 @@ msgstr "Silenciar la salida de sonido"
 msgid "Disables output of sound"
 msgstr "Desactivar la salida de sonido"
 
-#: src/snes9x.ui:4200
-msgid "Mute sound when using turbo"
+#: src/snes9x.ui:4206
+#, fuzzy
+msgid "Mute sound when using turbo or rewind"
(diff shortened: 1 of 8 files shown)


bearoso:
You don't need to update the .pos unless you're doing translation. 

I don't think you can just get rid of the mute when pausing. Some drivers will play garbage or just repeat their last buffer when that happens. There might be other safeguards preventing that now, but I'll have to check.

bearoso:
It works fine.

AlexApps99:
If it's needed, I could probably modify the code to store the previous state of the mute during pausing, to avoid these issues without potentially playing garbage

bearoso:
No, I think those are just a holdover from pre-1.52. It’s safe to remove them.

PR #654, opened 2020-07-28
auto select recent opened rom on GtkFileChooser

derekdai:


bearoso:
The way it works now is it shows the last directory opened through the "Open ROM" menu. With your change, if you select a recent item, it's moved to the top, and the directory it was in is selected. I think it's more consistent the way I have now.

derekdai:
The goal is let the dialog auto select the last opened ROM, this way, if I have a lot of ROMs in the same directory, I don't need to scroll long distance to fine next (relative to the last opened ROM) ROM.

PR #659, opened 2020-08-17
removed a unnecessary branch

Yusuf-EMU:


bearoso:
`time_ratio *= spc::dynamic_rate_multiplier * (Settings.DynamicRateControl);`
No, this doesn't work. If Settings.DynamicRateControl is 0 the whole thing is set to 0.

Yusuf-EMU:
K, fixed it

PR #667, opened 2020-09-12
remove signing from all targets

orsi:
Should remove all signing issues for Xcode projects


MichaelBuckley:
This change is will remove code signing from release builds, which will make Snes9x impossible to notarize for distribution to users. These changes appear to already be in place for debug builds, so you shouldn't need these changes to be able to build. But since you're still having problems with the debug builds, I'm going to merge this and just turn release signing back on temporarily every time I need to notarize a build for distribution.

PR #670, opened 2020-09-13
Adds frame skip stepper to preferences window

orsi:
Screen Shot 2020-09-12 at 10 03 12 PM

### DIFF
--- a/macosx/Snes9x/S9xPrefsConstants.m
+++ b/macosx/Snes9x/S9xPrefsConstants.m
@@ -25,4 +25,4 @@ NSString * const kJoypadInputPrefs = @"JoypadInputs";
 NSString * const kJoypadPlayerPrefs = @"JoypadPlayers";
 NSString * const kShowFPSPref = @"ShowFPS";
 NSString * const kVideoModePref = @"VideoMode";
-NSString * const kMacFrameSkipPref = @"MacFrameSkip";
+NSString * const kMacFrameSkipPref = @"FrameSkip";
--- a/macosx/Snes9x/S9xPrefsConstants.m
+++ b/macosx/Snes9x/S9xPrefsConstants.m
@@ -25,3 +25,4 @@ NSString * const kJoypadInputPrefs = @"JoypadInputs";
 NSString * const kJoypadPlayerPrefs = @"JoypadPlayers";
 NSString * const kShowFPSPref = @"ShowFPS";
 NSString * const kVideoModePref = @"VideoMode";
+NSString * const kMacFrameSkipPref = @"MacFrameSkip";
--- a/macosx/mac-render.mm
+++ b/macosx/mac-render.mm
@@ -297,6 +297,10 @@ void S9xPutImage (int width, int height)
         for (int i = 0; i < 60; i++)
             frameCalc += drawnFrames[i];
 
+		// avoid dividing by 0
+		if (frameCalc == 0)
+			frameCalc = 1;
+		
         IPPU.DisplayedRenderedFrameCount = (Memory.ROMFramesPerSecond * 60) / frameCalc;
     }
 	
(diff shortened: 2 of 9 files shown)


macosx/mac-render.mm
```diff
@@ -297,6 +297,10 @@ void S9xPutImage (int width, int height)
         for (int i = 0; i < 60; i++)
             frameCalc += drawnFrames[i];
 
+		// avoid dividing by 0
+		if (frameCalc == 0)
+			frameCalc = 1;
+		
         IPPU.DisplayedRenderedFrameCount = (Memory.ROMFramesPerSecond * 60) / frameCalc;
```
orsi, line 304: When macFrameSkip is set to 0, frameCalc will try to divide by 0 and throw an error here. I've set this as a placeholder for now, but I'm assuming there's a better way to do this?

MichaelBuckley: Yeah, that's an odd one.

I think the answer here is in S9xSyncSpeed(), instead of always setting `skipFrames` to `macFrameSkip`, we should set it to 1 when `macFrameSkip` is 0.

This code predates me, but it appears that `skipFrames` is only used to determine the emulated (but not necessarily drawn) frames per second. It should always be set to 1 when the frame was drawn without skipping, and set to the number of skipped frames + 1 otherwise.

MichaelBuckley:
Nicely done with the Objective-C. Aside from a few style differences, I don't see any issues, and I'm not super strict about style.

I would like to see a combo box here instead of a stepper, since that would allow us to add an explicit "Automatic" menu entry, so users don't have to know that they need to set it to -1.

I would feel good about merging this change unless you want to address anything else before I do so.

orsi:
I’ll take some time and change it to a combobox and see how it works. My initial thoughts were to have a separate toggle for automatic/fixed frame skip, which would disable the stepper when set to automatic.

I’ll play around with it a bit more and let you know.

macosx/Snes9x/S9xPrefsConstants.m
```diff
@@ -25,3 +25,4 @@ SNES9X for Mac OS (c) Copyright John Stiles
 NSString * const kJoypadPlayerPrefs = @"JoypadPlayers";
 NSString * const kShowFPSPref = @"ShowFPS";
 NSString * const kVideoModePref = @"VideoMode";
+NSString * const kMacFrameSkipPref = @"MacFrameSkip";
```
MichaelBuckley, line 28: One minor nitpick. You might just want the string here to be "FrameSkip". It's user visible if anyone ever looks in the preference plist file. Users don't really care if this is the frame skip variable the Mac port uses to keep track of things as opposed to the variable used in the cross-platform section of the code.

orsi:
No worries, I’ll fix it up!

bearoso:
Frame skip is kind of a pointless option at this point. Nothing can't keep up anymore, so the options you need are for how the program deals with refresh rate mismatches. So things like:
1. Keep full frame rate, skip a frame when behind a fixed timer.
2. Always display one frame per monitor refresh.
3. Try to display one frame per monitor refresh, duplicate a frame when the sound buffer is too full.

MichaelBuckley:
Unfortuantely, Apple still sells some fairly underpowered Macbook Airs that try to run high-resolution monitors (meaning we have to double the resolution) off of old i3s with Intel integrated graphics. They can keep up with snes9x as long as nothing else is happening on the system, but unfortunately things like spotlight indexing or checking for system updates in the background can slow the machine enough to affect emulation. Having the option at least for automatic frame skipping is kind of necessary for those machines.

orsi:
Tried the ComboBox, however I felt like it wasn't an ideal solution to have a single option (Automatic) and have to manually type in a number. I think the combination of a checkbox with the stepper/textfield number formatter is the best solution:

frameskip

This should be ready as long as everyone agrees on the ui.

MichaelBuckley:
Works for me. I'm going to merge it.

PR #671, opened 2020-09-14
Ignore osx files and remove cached/tracked .DS_Store

orsi:
.DS_Store not needed in project source


MichaelBuckley:
Thanks taking care of that annoyance.

PR #672, opened 2020-09-14
*WIP* - Redesign preferences window to macos ui style

orsi:
Work-in-progress -- pull request is for feedback and visibility. Seems to be working properly, however I haven't thoroughly tested yet.

macos-ui

Looking for more feedback in terms of layout, toolbar icons, and what preferences can be added back in (i.e. features are ready on MacOS and just need the ui to change them).

I could also add in UI elements for every feature which will eventually be included, but simply disable them for now so that in the future they can be easily flipped on.


macosx/Snes9x/AppDelegate.h
```diff
@@ -20,10 +20,17 @@
 
 #import <Cocoa/Cocoa.h>
 #import <snes9x_framework/snes9x_framework.h>
+#import "S9xPreferences/S9xPreferencesWindowController.h"
+
+static NSWindowFrameAutosaveName const kMainWindowIdentifier = @"s9xMainWindow";
```
MichaelBuckley, line 25: You typically want to have `extern NSWindowAutosaveName const` declarations in header files, and `NSWindowAutosaveName const` values in implementation files. Including the actual string in the header file creates copies in every file that includes the header, and if it's static, that can cause linking issues if the header file is included in multiple libraries.

So in this case, AppDelegate.h would read

`extern NSWindowFrameAutosaveName const kMainWindowIdentifier;`

And in AppDelegate.m:

`NSWindowFrameAutosaveName const kMainWindowIdentifier = @"s9xMainWindow";`

MichaelBuckley:
One thing we might want to be particularly careful about is documenting the images used, where they came from, and that we have the rights to use them.

Emulation is perfectly legal, but there are still companies that don't like it, and they occasionally look for ways to legally challenge emulators. It's probably overkill, but it's to our benefit to be as scrupulous as possible when it comes to legal issues.

macosx/Snes9x/AppDelegate.h
orsi: Ahh, didn't know that. I am guessing that's the same reason why a lot of C header files contain:

`#ifndef HEADER
#define HEADER
... declarations ...
#632`

It was split up that way before but I didn't realize why that was the case, so I'll try to revert the strings back the way they were.

orsi:
> One thing we might want to be particularly careful about is documenting the images used, where they came from, and that we have the rights to use them.
> 
> Emulation is perfectly legal, but there are still companies that don't like it, and they occasionally look for ways to legally challenge emulators. It's probably overkill, but it's to our benefit to be as scrupulous as possible when it comes to legal issues.
> 
> I'm specifically worried about the display icon. It appears to be one of Apple's copyrighted icons that we're copying into our image asset catalog rather than referencing the icon already present on the system.

I'll make sure to do credit and source the images better. I thought there wouldn't be an issue with using the Mac system icons, but perhaps I'm mistaken. The SNES controller/system were creative commons images on google, but perhaps I can look for more general images just in case.

What do you think is a good way to document the source of the images?

MichaelBuckley:
> I thought there wouldn't be an issue with using the Mac system icons, but perhaps I'm mistaken.

US Copyright law is _very_ weird. There's no problem with us writing code that references icons that are already on disk as part of the OS install. There's actually an API for that, though it's a bit obscure, requring you to know the 4-letter code of the image you want to use. Apple provides sample code that lists most of them. https://developer.apple.com/library/archive/samplecode/IconCollection/Introduction/Intro.html

But copying the images into our asset catalog can constitute copyright infringement.

> What do you think is a good way to document the source of the images?

Depends on the CC license the images are under. If they require attribution, we need to have a view in the app that lists all the attributions. This could be a link or a button in the about screen, or it could be its own menu under the application or help menus.

If the CC licenses don't require attribution, I think we can just document the sources in a git commit message.

macosx/Snes9x/AppDelegate.h
MichaelBuckley: Yes, it's related to the `#ifndef` `#define` header dance.

`#include` in C and C++ literally just copies and pastes the header in the location where the `#include` line is. The preprocessor does this before compilation starts. This is why there are some cases where the order of `#include`s matters.

The Objective-C `#import` does the `#ifndef` and `#define` automatically, but it still copies and pastes the contents of the header file.

orsi:
Just using the system icons and what is included in SNES9x already.

This is a good start I believe for the preferences window. Everything seems to be working properly.

preferences ui

macosx/Snes9x/AppDelegate.h
```diff
@@ -20,10 +20,17 @@
...
 #import <Cocoa/Cocoa.h>
 #import <snes9x_framework/snes9x_framework.h>
+#import "S9xPreferences/S9xPreferencesWindowController.h"
+
+extern NSWindowFrameAutosaveName const kMainWindowIdentifier;
 
 @interface AppDelegate : NSObject <NSApplicationDelegate, S9xInputDelegate>
 
 @property (nonatomic, readonly, assign) BOOL isRunningEmulation;
+@property (nonatomic, strong) S9xEngine *s9xEngine;
```
MichaelBuckley, line 30: I won't reject the change for this, but I'll probably move these back to the .m file. The purpose of defining properties in the .m file is for visibility. for properties, Objective-C doesn't have any notion of public/private/protected. Defining them in the .m file makes them private.

macosx/Snes9x/AppDelegate.m
```diff
@@ -42,25 +31,25 @@ - (void)applicationDidFinishLaunching:(NSNotification *)aNotification {
     [self setupDefaults];
     [self importRecentItems];
 
-    NSWindow *window = [[NSWindow alloc] initWithContentRect:s9xView.frame styleMask:NSWindowStyleMaskTitled|NSWindowStyleMaskClosable|NSWindowStyleMaskMiniaturizable|NSWindowStyleMaskResizable backing:NSBackingStoreBuffered defer:NO];
+    NSWindow *gameWindow = [[NSWindow alloc] initWithContentRect:s9xView.frame styleMask:NSWindowStyleMaskTitled|NSWindowStyleMaskClosable|NSWindowStyleMaskMiniaturizable|NSWindowStyleMaskResizable backing:NSBackingStoreBuffered defer:NO];
```
MichaelBuckley, line 34: Good rename.

MichaelBuckley (approved):

macosx/Snes9x/AppDelegate.h
orsi: Please do what you think is best. I shouldn't have moved them around anyways. I think I was just moving some variables around to get an idea of how Objective-C works and whether declaring properties in the .h or .m file would break the compiler.

macosx/Snes9x/AppDelegate.m
orsi: Intended to differentiate from the preferences window now.

macosx/Snes9x/AppDelegate.h
MichaelBuckley: Yeah, no worries on that front. It certainly seemed like something like that was going on, and I just wanted to provide some context that isn't often called out in Objective-C documentation.

PR #682, opened 2020-12-10
Initial code to add Flipflop functionality; barely tested

cptnoremac:
Add a Flipflop function to the Input menu. It should swap the controls between player 1 and player 2 every few seconds with a random interval. It should also play a countdown sound to warn you when it's about to switch.


cptnoremac:
Whoops. I meant to PR this in my own repo.

PR #711, opened 2021-07-31
Add proper support for SuperFX 8MB ROM emulation

Yoshimaster96:
This code adds support for SuperFX ROMS which use the extended 6MB CPU ROM region specified by the official SNES documentation. It's not super well-tested though.


Yoshimaster96:
Please note that the 4MB ROM size mapping might not be 100% accurate (specifically with banks $E0-$FF). I'm not sure if this is mirrored or if it's open-bus. It shouldn't be too big an issue though.

qwertymodo:
It looks like either your git config or your text editor converted the EOL format on the files, causing the entire file to register as changed.  This makes it very difficult to actually understand what you've changed.  Please fix that and resubmit.

Yoshimaster96:
Accidentally closed and now some checks don't work...

Yoshimaster96:
Oh okay, hold on...

Yoshimaster96:
There we go, that should work.

qwertymodo:
Also, I'm not familiar with this expanded ROM region.  Where did you find it specified?  I'm guessing this must be cartridge ROM mapped directly to the main A bus, since the SuperFX is only capable of mapping up to 2MB on the GSU-2 revisions and 1MB on everything else.  So, this sounds really similar to the Tengai Makyou Zero translation mapping, where there's an extra 1MB ROM mapped directly to the A bus along with the 4MB ROM mapped to the SPC7110, is that right?

qwertymodo:
That didn't work, because now it just shows the entire files changed twice.  Try squashing the commits into one and force pushing.

PR #712, opened 2021-07-31
Add proper support for SuperFX 8MB ROM emulation

Yoshimaster96:
This code adds support for SuperFX ROMS which use the extended 6MB CPU ROM region specified by the official SNES documentation. It's not super well-tested though.


Yoshimaster96:
Deleted repo and reuploaded from scratch.

qwertymodo:
To copy my question over from the previous submission, where exactly is this expanded ROM mapping documented?  Is this just mapped directly to the main CPU A bus in the same way as the Tengai Makyou Zero english translation does with its extra 1MB ROM?  The SuperFX itself is only capable of mapping a maximum of 2MB of ROM in the GSU-2 revisions, and only 1MB of ROM in all other revisions.

Yoshimaster96:
It's in the official SNES Development Manual Book 2.

qwertymodo:
Ok, thanks for that.  If I'm reading your code correctly, the SuperFX ROM is the first 2MB of the file, and the CPU ROM is everything after that?  Are you only supporting files that are exactly 8MB in size, or does this also handle say a 4MB file, with 2MB of SFX ROM and 2MB of CPU ROM?

Yoshimaster96:
I tried to handle 4MB ROMs, but I dunno how exactly they're mapped. The way I implemented it was to mirror the extra 2MB into the remaining memory, but that might not be 100% accurate.

bearoso:
What's with the extra mapping
```
		map_space(0xf0, 0xf0, 0x0000, 0xffff, SRAM);
		map_space(0xf1, 0xf1, 0x0000, 0xffff, SRAM + 0x10000);
```
added to the classic 2MB case?

Yoshimaster96:
Haven't worked on this stuff in forever, so I don't remember exactly why that was added. It does seem to be the behavior used by bsnes-plus however.

niuus:
> Haven't worked on this stuff in forever, so I don't remember exactly why that was added. It does seem to be the behavior used by bsnes-plus however.

Hi @Yoshimaster96. I've reading about Eggflutter while looking for example hacks that use the Super FX 8MB ROM, in order to test it in [Snes9xRX](https://github.com/niuus/Snes9xRX/). I resorted to just opening and re-saving 2 hacks with Eggflutter ([Comeback to Yoshi's Island](https://www.smwcentral.net/?p=section&a=details&id=20980) / [SMWC Island](https://www.smwcentral.net/?p=section&a=details&id=30281)) just to see if they could at least run, which they do in both Snes9x 1.62.2 and Snes9xRX, even if they hang when about to start a level. However, I stumbled upon this [post](https://www.smwcentral.net/?p=viewthread&t=118850&page=2):

"So it seems I may have made a major mistake in the v1.0x patch.

Apparently accessing banks $70 and $71 via the ROM pointer is not allowed. bsnes just happens to allow this, which is why it wasn't noticed before. However the Snes9X modification I'm working on correctly does not allow this, and so hacks made with v1.0x won't work on it.

This will be fixed in the v1.1x patch however along with the newly added features, so don't worry. Hacks opened up in the upcoming 1.1x version will be automatically patched to correct the issue."

Was this addressed in Eggflutter? I couldn't find any version past 1.02.

Yoshimaster96:
I think so? I'm not entirely sure. Like I said, I haven't worked on this stuff in a LONG time.

niuus:
> I think so? I'm not entirely sure. Like I said, I haven't worked on this stuff in a LONG time.

Okey. I take it 1.1 was unreleased and will remain so permanently? It's okey if it won't, i'll just leave it as-is and close the chapter.

Yoshimaster96:
Yeah it probably won't release for quite a while if ever. Sorry about that.

PR #717, opened 2021-08-24
Fix Win10 SDK and v142 toolset build

ImmaturityRules:
This is a small modification to allow snes9x to build on Windows 10 with Visual Studio 2019, the latest Windows 10 SDK and the v142 C++ toolset. The DirectX June 2010 SDK is still needed because of the known issues with XAudio2.

All it does is undefine some symbols to prevent redefinition errors when including the Win10 SDK and DXSDK headers, and conditionally prevent the inclusion of wincodec.h if WINVER is not below Windows 10 because some errors are generated if it's included. The changes are isolated to the dxerr.cpp compilation unit, I believe.

I've tested the changes on the recommended build configuration (Windows 7, VS2017 with the DXSDK and v141_xp toolset) and nothing seems to be broken. The executable runs without errors and I loaded a ROM without problems.


OV2:
Thanks.
The dx sdk dependency could probably be removed with a custom xaudio2 header, but I never really had a need for it.

PR #735, opened 2022-02-04
Microoptimisations

jSTE0:
A few microoptimisations noticed during code inspection.


jSTE0:
Would it be easier to get this merged if I broke it into multiple PRs? I've tested them on Fedora 35 x86-64 with gcc-11.2.1 and a downstream fork of an earlier version of snes9x (see the PR that references this one) on ARMv5TE with gcc-11.2.
Least controversial:
- Patch 1: The performance issue with IPPU.FrameCount seems to have existed as far back as snes9x-1.34a (the earliest I could find source for). Saves a few tens of clock cycles so possibly low thousands per second, still a tiny amount even on old sub-GHz processors.
- Patch 2: This appears to be a performance regression introduced in snes9x-1.52 when the const arrays moved from global to function scope.

Less controversial would be patches 4 and 5: The compiler (on the tested LE platforms) does recognise the pattern and optimises it as a single 2B read/write anyway but somehow it does still lead to slightly shorter assembler code, perhaps by doing the optimisation "by hand" it allows the compiler to recognise other optimisations. In any case, it may make the code more readable.
- Patch 4: In `fx_readRegisterSpace()` the use of `+=` means the compiler doesn't recognise the pattern and uses a byte-by-byte so this a win.
- Patch 5: The long-standing bug with `Op11Xr` doesn't seem to affect the few titles I tried as debug output shows that it's 0 anyway.

For patch 3, the gcc built-in function should compile down to a single instructions on the platforms that this is likely to be deployed on: x86 (bsr), ARMv5 (clz), PowerPC (cntlz), MIPS (clz). Latency is 1 clock cycle on an ARM926EJ-S, 7-73 clock cycles on the original P5 Pentium to 10 cycles on K7 Athlon and 3-4 on modern x86 CPUs. On CPUs with ABM (K10 Phenom) or BMI1 (Haswell), lzcnt can be used with latency of 1-2 clock cycles if the binary is built for the appropriate CPU.
Hope this helps.

bearoso:
I'm hesitant about the macro and intrinsics in patch 3, but the DSP code is an ancient disaster, so whatever.

jSTE0:
Thanks for merging this. I did consider adding an MSVC codepath as that uses a different intrinsic but that'd make the code even messier (perhaps factor it out into `port.h`?) but more importantly I don't have that compiler to test it with.
I'll try to submit backports of the appropriate patches to the various libretro forks soon.

PR #746, opened 2022-03-07
Fix Rockman X 1.0 memory mapping to stop triggering the botched copy …

AkiteruSDA:
…protection

Cartridges of the 1.0 version of Rockman X have bodge wiring to disable some ROM mirroring due to faulty copy protection being triggered when it shouldn't, which is described [here](https://web.archive.org/web/20210127154007/https://byuu.net/cartridges/boards/) better than I could ever explain it.

Before snes9x 1.57, this wasn't an issue because CMemory::map_LoROMSRAM still ran when opening LoROMs with SRAM size 0. However, [this commit](https://github.com/snes9xgit/snes9x/commit/386bfe0aa9c33b33d84e941f0d2055830dde3d00) stopped mapping SRAM in that case, which re-enabled those copy protection issues.

The reason for adding CMemory::Map_RockmanXRev0LoROMMap was that this is a special hardware case that modifies the behaviour of the stock board, and other special carts such as the Sufami Turbo have their own functions as well. If anyone has any recommendations on how to implement this functionality that may be preferable, please let me know.

Thanks!

### DIFF
--- a/memmap.cpp
+++ b/memmap.cpp
@@ -2508,9 +2508,6 @@ void CMemory::InitROM (void)
 			}
 		}
 		else
-		if (strncmp(ROMName, "ROCKMAN X", 9) == 0 && ROMVersion == 0)
-			Map_RockmanXRev0LoROMMap();
-		else
 			Map_LoROMMap();
     }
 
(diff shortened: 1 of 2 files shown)


bearoso:
I don't remember why I changed this in the first place. Anyone else know?

qwertymodo:
I have a feeling that this might have been due to me adding 128KB SRAM mapping support.  I don't know if this really warrants a game-specific hack like this, there might be a cleaner solution based on the SRAM size header byte.

bearoso:
I'm just going to revert the original commit for now.

AkiteruSDA:
Thanks for the feedback. I removed the new mapping type and instead mapped just the necessary SRAM addresses as a case in CMemory::ApplyROMFixes. That seems like a more appropriate place for that code given the nature of the issue.

bearoso:
This should be working properly in the latest master.

PR #795, opened 2022-11-01
Fixes issue where non-archive msu-1 files would have improper file extensions

kbrighton:
I recently noticed an issue where SNES9x was not properly loading MSU-1 files on GTK Linux.  After stepping through the code, I found out that it was due to the fact that it was attempting to open <romname>.-1.pcm rather than the correct <romname>-1.pcm.

This should correct the behavior with regards to the PCM files while keeping the functionality as it was for the non PCM files.

### DIFF
--- a/gtk/src/gtk_file.cpp
+++ b/gtk/src/gtk_file.cpp
@@ -121,8 +121,15 @@ const char *S9xGetFilename(const char *ex, enum s9x_getdirtype dirtype)
     static std::string filename;
     fs::path path(S9xGetDirectory(dirtype));
     path /= fs::path(Memory.ROMFilename).filename();
-    path.replace_extension(ex);
-    filename = path.string();
+    //Fixes issue with MSU-1 on linux
+    if(ex[0] == '-') {
+        path.replace_extension("");
+        filename = path.string();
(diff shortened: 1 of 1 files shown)


bearoso:
The MSU1 code is actually abusing undefined behavior in the S9xGetFilename function. It should probably be fixed at that spot. 

Coincidentally, I actually rewrote that in my private branch.

kbrighton:
Thinking on it, you're probably right, especially if it can be achieved in a way that's correct for all the implementations of S9xGetFilename.  I fixed it like this because it appeared to have been an issue specifically with how the gtk version was implemented, as it used the fs::path functions instead of the older ways of doing so.

Honestly, I'd love to see your fix of it, if it's more universal than what I was doing.

qwertymodo:
I recall having a similar issue with the Wii port, which didn't implement S9xGetFilename at all. I initially implemented MSU-1 on the Windows port, and the behavior of S9xGetFilename as it's implemented there, on the (incorrect, it would seem) assumption that the behavior was the same across platforms.

bearoso:
I'm merging now since this is probably the easiest temporary fix.

PR #808, opened 2023-01-10
macOS: Set LSApplicationCategoryType to games

UltraHDR:
https://developer.apple.com/documentation/bundleresources/information_property_list/lsapplicationcategorytype macOS will auto-add it to the agmes folder in macOS Launchpad


MichaelBuckley:
Merged. For anyone following along, this is a change that only affects a resouce file used by macOS to query for information about the app. There are no code changes, and no negative side-effects of adding this value to the resource file.

PR #813, opened 2023-02-04
Fix dvision routine on SA-1 returns the wrong reminder (and quotient) on negative dividends

### LINKED ISSUE #799: Division routine on SA-1 returns the wrong reminder (and quotient) on negative dividends.
The SA-1 division routine that uses the $2254+ registers returns the wrong values for reminder and quotient when given negative dividends.

The code in question: https://github.com/snes9xgit/snes9x/blob/master/sa1.cpp#L564-L568
For [example](https://godbolt.org/z/4e9hWzb8e), when given inputs of -10 / 5, it returns 5 as remainder and 65533 as quotient. 

As seen from the linked example code, BSNES returns the correct result: https://github.com/bsnes-emu/bsnes/blob/master/bsnes/sfc/coprocessor/sa1/io.cpp#L434-L442.

The fix is either to add the missing `% divisor` at the end of the reminder calc or just replace the routine with this slightly improved code with avoids a couple of modulos (verified that it returns the correct result for every possible input pair):
```c++
    uint32 dividend_ext = dividend + (uint32)divisor*65536;
    uint16 remainder = dividend_ext % divisor;
    uint16 quotient = dividend_ext / divisor;
```
Improved code kindly provided by @Alcaro .
Credits to @Fernap for finding the issue after debugging an SA-1 sprite which was crashing the game.

Atari2:
This PR fixes #799 
For more details read the issue's description.

### DIFF
--- a/sa1.cpp
+++ b/sa1.cpp
@@ -563,8 +563,9 @@ void S9xSetSA1 (uint8 byte, uint32 address)
 					{
 						int16 dividend = (int16) SA1.op1;
 						uint16 divisor = (uint16) SA1.op2;
-						uint16	remainder = (dividend >= 0) ? dividend % divisor : (dividend % divisor) + divisor;
-						uint16	quotient  = (dividend - remainder) / divisor;
+						uint32 dividend_ext = dividend + (uint32)divisor * 65536;
+						uint16 remainder = dividend_ext % divisor;
+						uint16 quotient = dividend_ext / divisor;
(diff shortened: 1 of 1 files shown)


Atari2:
The failed checks are because of: https://cirrus-ci.org/blog/2022/11/08/sunsetting-intel-macos-instances/

bearoso:
Ugh. This isn't the first time we've messed with this. I hope it's right this time.

PR #851, opened 2023-05-21
q -> a typo fix

jeffythedragonslayer:


bearoso:
A word of warning. Do not create pull requests for something as minuscule as a typo in an unnecessary comment. If you went and fixed *all* the typos, I *might* think about merging it, but this is just blatant abuse.

PR #852, opened 2023-05-30
Support Nimbus+ controller in mac-joypad.mm findControls()

socantre:
The findControls() routine was not finding several of the buttons on my Nimbus+ controller so that I could not configure those buttons in

    Settings > Controls > Nimbus+

A, B, X, Y, and the L and R shoulder buttons are detected by the existing code. The D-pad and the 'start' and 'select' standins are not.

One odd thing that happened the first time I played with this change was that after configuring the new buttons and loading up a game the buttons seemed to be initially in a random state and so the game seemed to behave like some random buttons were held down. After pressing and releasing all the new buttons the game seemed to behave more normally. I'm not sure of my diagnosis because it only happened the one time and I didn't have an opportunity to reproduce or debug whatever happened.

### DIFF
--- a/macosx/mac-joypad.mm
+++ b/macosx/mac-joypad.mm
@@ -282,7 +282,9 @@ void findControls(struct JoypadDevice &device, NSDictionary *properties, NSMutab
 
     int usagePage = [properties[@kIOHIDElementUsagePageKey] intValue];
     int usage = [properties[@kIOHIDElementUsageKey] intValue];
-    if (usagePage == kHIDPage_Button)
+    if (usagePage == kHIDPage_Button ||
+        (usagePage == kHIDPage_GenericDesktop && (usage == kHIDUsage_GD_DPadUp || usage == kHIDUsage_GD_DPadDown || usage == kHIDUsage_GD_DPadLeft || usage == kHIDUsage_GD_DPadRight))
+        || (usagePage == kHIDPage_Consumer && (usage == kHIDUsage_Csmr_ACProperties || usage == kHIDUsage_Csmr_ACExit)))
     {
         [buttons addObject:properties];
     }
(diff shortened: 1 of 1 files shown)


macosx/mac-joypad.mm
```diff
@@ -282,7 +282,9 @@ void findControls(struct JoypadDevice &device, NSDictionary *properties, NSMutab
 
     int usagePage = [properties[@kIOHIDElementUsagePageKey] intValue];
     int usage = [properties[@kIOHIDElementUsageKey] intValue];
-    if (usagePage == kHIDPage_Button)
+    if (usagePage == kHIDPage_Button ||
+        (usagePage == kHIDPage_GenericDesktop && (usage == kHIDUsage_GD_DPadUp || usage == kHIDUsage_GD_DPadDown || usage == kHIDUsage_GD_DPadLeft || usage == kHIDUsage_GD_DPadRight))
+        || (usagePage == kHIDPage_Consumer && (usage == kHIDUsage_Csmr_ACProperties || usage == kHIDUsage_Csmr_ACExit)))
```
MichaelBuckley, line 287: Adding support for kHIDUsage_GD_DPad* makes sense. SDL checks for these (and a few others besides, which I should probably add).

The kHIDUsage_Csmr*s, however, are too specific. It took me a while to figure out why your controller wanted these. If you look at the USB HID spec, these are defined as buttons for controlling GUI applications. ACProperties is "Display the properties of the current document" and ACExit is "Exit the application"

But a comment in the SDL source code says that it looks at the entire kHIDPage_Consumer "e.g. 'pause' button on Steelseries MFi gamepads."

It sucks that gamepad manufacturers are using this page, but there's not much we can do about that. But it's probably best to just check usagePage == kHIDPage_Consumer and remove the usage checks for kHIDUsage_Csmr_ACProperties and kHIDUsage_Csmr_ACExit.

MichaelBuckley:
@socantre Thank you for the patch and for the revision!

PR #875, opened 2023-09-23
Fix GCC 13 build issue

spillner:
Prevent g++ from complaining that snprintf() is undefined when VMA_STATS_STRING_ENABLED is set.


bearoso:
Thanks, but I'll just update to the new release of Vulkan Memory Allocator.

PR #900, opened 2024-01-28
Updated Ukrainian translation

StanleyKid-22:
Updated to actual version 1.62.3.
The [previous request](https://github.com/snes9xgit/snes9x/pull/899) was done in a hurry - new commit pushing didn't work, so I just closed and created a new one.
_I apologize for that inconvenience._


bearoso:
OpenGL should just remain "OpenGL" because it's a name and trademark. Same goes for Vulkan.  Also, you don't need to translate the license, and probably shouldn't, because it might change some of the wording and the meaning.

StanleyKid-22:
OK, got it. I'll keep in mind in the future.
About the license text and not only - maybe then it makes sense to remove from the POT file those strings that definitely won't need translation?

PR #950, opened 2024-09-14
Fix load with preview file names and time stamps on extra banks

### LINKED ISSUE #949: Load with preview dialog timestamps incorrect
The "Load with preview" dialog is showing the wrong filenames and timestamps under my saves when I use multiple banks.
When I switch from bank 0 to bank 1, the preview screenshots change, but the text under them doesn't. Or well, I think it does get redrawn, but the source of the information seems to be wrong. Most of my saves were created with 1.62.3, but even newly created saves with 1.63 are affected.

As a side note, the dialog doesn't seem to play well with UI scaling in Windows 10. Mine is set to 150% because 4k at 27" can get a bit small otherwise, but that makes the timestamps hard to read to begin with → on second thought, this is probably because of the long ROM file name causing wrapping.

**For example:**
snes9x-load-with-preview

**Affected versions:** 1.63, 1.62.3

**Analysis:**
Unfortunately my C++ is extremely rusty, so I don't think I'd be doing you a favor with a PR. However, this code looks to be incorrect:
https://github.com/snes9xgit/snes9x/blob/0c547f34868639d4f4d0de5dff2cae6a42007c67/win32/CSaveLoadWithPreviewDlg.cpp#L129
It does not even take the `bank` as a parameter, so there is no way it could generate the right file name. This means it will show the info for the same slot in bank 0 always, unless that slot isn't in use on bank 0, in which case I think it will only show the (incorrect) filename.

Thorarin:
Fixes #949

### DIFF
--- a/win32/CSaveLoadWithPreviewDlg.cpp
+++ b/win32/CSaveLoadWithPreviewDlg.cpp
@@ -126,7 +126,7 @@ void CSaveLoadWithPreviewDlg::load_slot_image_text(int slot, int bank, HWND hDlg
         SendMessage(GetDlgItem(hDlg, IDC_BUTTON_SLOT_1 + slot), BM_SETIMAGE, (WPARAM)IMAGE_BITMAP, (LPARAM)previewHbmps[slot]);
 
         char filename[_MAX_PATH + 1];
-        GetSlotFilename(slot, filename);
+        GetSlotFilename(bank * SAVE_SLOTS_PER_BANK + slot, filename);
         Utf8ToWide filenameW(filename);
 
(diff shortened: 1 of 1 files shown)


Thorarin:
The Windows build is broken because of this commit, by the way:
https://github.com/snes9xgit/snes9x/commit/ea243051edbb7a7882b37d967b4f87149e1f9ff8
I wasn't able to find an easy replacement for `localtime_r` function, so I just tested locally with this particular line commented out and left it unchanged.

bearoso:
Looks fine. Merged.

PR #966, opened 2024-12-17
Update main.cpp: fixed application name

gilchavezm:
Fixed desktop application name


bearoso:
This is temporary and intentional to get desktops to pick up the XDG icon for snes9x-gtk while there's no .desktop file for qt yet.

PR #986, opened 2025-07-22
New Interpolation Filters

AltoWildgruber:
I wrote three interpolation filters that can be easily added to the filter dropdown list by including them in render.cpp and wsnes9x.h. I put the filters into the filter folder.

1. Bicubic 4x: This is a gamma-corrected bicubic upscaler based on Hermite. It's sharper than standard bilinear filtering and causes less smearing.

2. Lanczos 4x: It's a 4x upscaling filter based on the Lanczos algorithm. The output is really sharp. I reduced the bold outlines - a common side effect of blending in RGB-color space - by reducing the influence of true black pixels to blending.   

3. Sharp Bilinear (4x): It's a gamma-corrected bilinear filter that looks better than the built-in bilinear filter as it does not cause overly bold outlines. This is achieved by blending in gamma-space instead of in RGB. The filter is really fast. It uses a smoothstep variation of the standard bilinear algorithm which makes it considerably sharper and less blurry than standard bilinear.

### DIFF
--- a/filter/filter_bicubic_win.cpp
+++ b/filter/filter_bicubic_win.cpp
@@ -13,7 +13,7 @@ static uint8_t gamma_decode[256];
 
 static void init_gamma_tables()
 {
-    constexpr float gamma = 1.55f;
+    constexpr float gamma = 1.8f;
     constexpr float inv_gamma = 1.0f / gamma;
 
     for (int i = 0; i < 32; ++i)
--- a/filter/filter_bicubic_win.cpp
+++ b/filter/filter_bicubic_win.cpp
@@ -13,7 +13,7 @@ static uint8_t gamma_decode[256];
 
 static void init_gamma_tables()
 {
-    constexpr float gamma = 1.5f;
+    constexpr float gamma = 1.45f;
     constexpr float inv_gamma = 1.0f / gamma;
 
     for (int i = 0; i < 32; ++i)
@@ -28,7 +28,7 @@ inline int hermite_weight(int t_fp)
 {
     int t = (t_fp > FP_ONE) ? FP_ONE : t_fp;
     int inv = FP_ONE - t;
-    return (inv * inv) >> FP_SHIFT; // Full quadratic
+    return ((inv * inv * (FP_ONE + (t >> 1))) >> FP_SHIFT) >> FP_SHIFT; // Softer (approx. 2/3)
 }
 
 inline uint16_t build_rgb565_fast(int r, int g, int b)
--- a/filter/filter_bicubic_win.cpp
+++ b/filter/filter_bicubic_win.cpp
@@ -13,7 +13,7 @@ static uint8_t gamma_decode[256];
 
 static void init_gamma_tables()
 {
-    constexpr float gamma = 1.45f;
+    constexpr float gamma = 1.55f;
     constexpr float inv_gamma = 1.0f / gamma;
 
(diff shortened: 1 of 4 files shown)


AltoWildgruber:
@OV2 
Great! Try Sharp Bilinear first, it turned out to be the best and makes bicubic4x kind of unneccessary!
Lanczos 4x is only for those wanting an extremely crisp output (quite dissimilar to what CRTs did).

OV2:
I could see that Sharp Bilinear might appeal to some, although it is very slow. Seems to be fast enough in release.
Bicubic looks like a slightly darker version.
Laczos doesn't really provide any benefit over simple4x in my opinion.

AltoWildgruber:
**Overall you can say:**:

**CRT**: enlarges the native 224p image of an snes (analog) by scanline drawing: Does not lead to pixelation or blockiness, but to visible scanlines and slight blur or glow.

**Simple4x:** enlarges the original image using a simple nearest neighbor algorithm: the output is sharp but blocky and pixelated (not similar to CRTs)

**Bilinear Interpolation**: Leads to a properly upscaled picture that is not pixelated or blocky, but can be too blurry (more blur than CRTs added)

**Sharp Bilinear**: Aims to improve upon bilinear by making it sharper. Probably the closest to the original picture on a CRT, but without scanlines (can be added by loading a shader)

**Lanczos** is one of the most advanced and sharpest upscaling algorithms out there. But it may not be neccessary for SNES games. Although I personally think that some games benefit from the sharp output (e.g. Mario World).

AltoWildgruber:
In some games you can notice the difference. simple4x (quadruples pixels =Nearest Neighbor scaling), lanczos: sharp upscaling. For example in Yoshi's Island, you can see that Yoshi's outlines are less pixelated with lanczos when compared to simple 4x. But if you don't like the Lanczos filter, no problem. You don't have to take. It's also not as fast as I wanted it to be.

AltoWildgruber:
Also you have to take my updated Lanczos filter with radius 1.0. Earlier the radius was too low to see any difference.

AltoWildgruber:
Simple4x (Nearest Neighbor)

Lanczos

AltoWildgruber:
The first Donkey Komg screenshot is simple4x
The second lanczos.
I know that the difference is small: you basically have less pixelation at edges.
But when compared to other upscaling algorithms it's generally regarded the sharpest.

AltoWildgruber:
I could increase context float radius further to 1.5, which would make it clearer that the picture is indeed interpolated, but I would also lose sharpness. Overall Nearest Neighbor (simple4x) will always be the sharpest, unfortunately NN can't be matched in this regard.

AltoWildgruber:
I replaced sharp bilinear with a version optimized for performance.

OV2:
For now I've added the sharp bilinear algorithm. The high-res modes would need an adjusted version, but it's probably not worth it.

AltoWildgruber:
Great!

AltoWildgruber:
I don't know if you get feedback from the community, but if people like sharp bilinear, but would also like a sharper upscaler, then I think I can make a sharper variant of Sharp Bilinear. 

Lanczos is probably not the perfect solution, as it can be slow on some systems. 

Overall, the slight blur that the current sharp bilinear filter still comes with can be beneficial in games such as Donkey Kong country as it leads to smoother transitions.

PR #1022, opened 2026-02-02
Fix button state for HAT switch D-Pad

itsoli:
Fix input state for gamepads that report their D-Pad as HAT switch (`kHIDUsage_GD_Hatswitch`). Reliably handles mapping values of common logical ranges to the respective cardinal direction or the combination thereof (e.g 0-7, 0-8, 1-8, 0-3, 1-16, etc).

### DIFF
--- a/macosx/mac-joypad.mm
+++ b/macosx/mac-joypad.mm
@@ -34,6 +34,8 @@
 #include "mac-os.h"
 #include "mac-joypad.h"
 
+static const int32 CARDINAL_DIR_COUNT = 4;
+
 std::unordered_set<JoypadDevice> allDevices;
 std::unordered_map<JoypadDevice, std::map<uint8, std::map<int8, S9xButtonCode>>> defaultAxes;
 std::unordered_map<JoypadDevice, std::map<uint8, S9xButtonCode>> defaultButtons;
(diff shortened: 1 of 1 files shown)


macosx/mac-joypad.mm
```diff
@@ -220,43 +222,55 @@ void gamepadAction(void *inContext, IOReturn inResult, void *inSender, IOHIDValu
...
-					inputStruct.value = button;
-					if (buttonCodeByJoypadInput.find(inputStruct) != buttonCodeByJoypadInput.end())
-					{
-						pressedGamepadButtons[playerNum][buttonCodeByJoypadInput[inputStruct]] = true;
-					}
-				}
-			}
+            if (value >= logicalMin && value <= logicalMax)
+            {
+                int32 value0 = value - logicalMin;
```
MichaelBuckley: First of all, thank you so much for contributing this change. I think it's going to fix a lot of the issues that have been reported over the years with certain controllers that I couldn't get access to. All the controllers I have with hat switches work with the code form before this change, so it was frustrating that I couldn't figure out what was wrong.

I'm going to approve this pull request, but I would like to rename these variables to make them more descriptive. Names like `value0`, `dir1`, etc. make it harder to understand what they're for, at least to me.

I'll leave the pull request open for a couple days in case you'd like to name the variables yourself. Otherwise, I'll merge it in and then choose different names for them.

MichaelBuckley: Again, thank you so much. I'll merge this and try to come up with some better names, but you're right. It's difficult.

PR #1031, opened 2026-03-14
Added cheat code group editing. fix error C3688 when compiling snes9x_imgui.cpp

jiangsheng:
Changes max cheat size from 1027 to 32767 (The CHEAT_SIZE constant only affect the Windows build, though, other platforms may have different size limit). 

Add a group edit button to cheat code dialog.

Take the current cheat code to another dialog for editing. The dialog allows for multiline editing of cheat codes. 

Generate a line of '+' delimited cheat code from the lines when done. The engine would recognize the  '+' delimited line as a cheat group. 

Set the content of the cheat code textbox in the cheat code dialog if the user confirms the code change.

Also Visual Studio mistake snes9x_imgui.cpp as ASCII, force UTF-8 when compiling.


OV2:
I've fixed the problem with character count and reworked the wstring buffer handling a bit. Thank you for adding the dialog!

PR #1034, opened 2026-04-03
Add Visual Studio 2022 (v143) build support

shanytc:
- Retarget all vcxproj files from v141_xp/v141 to v143 toolset
- Update WindowsTargetPlatformVersion from 7.0/8.1 to 10.0
- Remove missing inffas8664.c reference from zlib (removed in latest zlib)
- Fix _tfwopen.h: remove non-standard _Prot parameter from ifstream constructors (removed in MSVC v143)
- Fix XAudio2 compatibility: replace deprecated GetDeviceCount/ GetDeviceDetails/XAUDIO2_DEVICE_DETAILS (XAudio2.7) with XAudio2.8+ API
- Fix dxerr.cpp: include dxgi.h, d3d10_1.h, d2d1.h, dwrite.h before dxerr.h to resolve undeclared DXGI/D2D/DWrite error identifiers


OV2:
The current project already builds in studio 2022 out of the box if you install "msvc v141 - vs 2017 c++ x64/x86 build tools" in the studio 2022 installer.

Xaudio 2.8 would need a GetDeviceCount/ GetDeviceDetails replacement for output device selection (maybe the waveout multimedia device enumeration could be used).

shanytc:
> > this is because v143 is a superset of v141 hench why the toggle for those who still defaults to xp (why would they anyway?)
> 
> You'd be surprised the number of people who will complain if you remove XP support.

Interesting... well, the PR exists, with a .ps1 shell to switch between v141/v143 for lazyness. or it can be closed for future reference :)

PR #1036, opened 2026-04-04
Native Kaillera Netplay Support

shanytc:
This PR introdces Kaillera native build support (No .dll) using the Kaillera Protocol and SDK for Windows Build.
Included:
- Host Kaillera Server
- Kaillera Client

Netplay Menu
image

Kaillera Server
image

Kaillera Client
image


bearoso:
We won't integrate Kaillera officially. The license is incompatible, and the project is old and hostile.. You'll have to keep this as something separate.

Also, LLM code is a licensing gray area that we don't have a policy for at the moment. At a quick glance, I don't see anything handled incorrectly, but I'd have to be sure to review it more thoroughly than manually written code.

bearoso:
I've followed along with your repository, and here's some additional insights:

Aside from Kaillera being unmergable because of licensing, your initial work looked ok, with some minor problems. For example, the Qt port was designed to be agnostic to the backend, with separation of concerns. Your patches pulled some snes9x code into the GUI part. This stuff is easily fixable. We'd also want to avoid including into the tree all the excess Retroachievements library code and pre built DLLs.

The newer stuff like the tile viewer are kind of dangerous, though. It duplicates a lot of the base graphics code into the win32 port and unrolls it. It manipulates the core data, potentially causing interactions, and assumes an unchanging API/ABI. This kind of feature needs to be built on an abstraction of the core drawing code to avoid adding technical debt. I'm also not a fan of the way it draws to the window using win32 bitblts. It's very slap-dashed and brute-forced the way LLMs generally tend to go.

The run ahead implementation is bad, hacky. I know Snes9x isn't properly instanced, but there's much safer ways to do it. I've avoided adding this feature because we have so many latency reduction tricks now that I don't think it's necessary. Super Mario World is the poster-child because of its built-in two frame delay, but it's not as significant most of the time. To do run-ahead correctly, we need instancing, which the Qt port is leaning towards. Retroarch is able to load multiple copies of the code to get around it. That avoids the kind of hacks your implementation does, but it's also not the best solution.

shanytc:
@bearoso 
I will address all the issues in due time. As for Kaillera, im not sure what is the issue with the licensing? I've implemented the Kaillera client from scratch using sdk. Do you mean that the project itself doesn't come with licensing hence why you do not want to include anything related to kaillera? (https://github.com/kaillerareborn/kaillerareborn.github.io)

PR #1049, opened 2026-05-01
Added support for a 11MB SuperFX mapper, and 256/512/896KB SRAM

dargereldren:
Tested to not break any existing games:

* Dirt Trax FX (Europe).sfc
* Dirt Trax FX (USA).sfc
* Doom (USA).sfc
* Star Fox - Super Weekend (USA).sfc
* Star Fox (USA).sfc
* Stunt Race FX (E).smc
* Stunt Race FX (Europe).sfc
* Stunt Race FX (U).smc
* Super Mario World 2 - Yoshi's Island (USA).sfc
* Vortex (USA) (En,Es).sfc
* Wild Trax (J).smc
* Winter Gold (Europe).sfc

Let me know if any issues arise in these though, but it shouldn't, since they all have proper ROM headers. This also extends MAX_ROM_SIZE to 0x1000000 to allow 16MB ROMs to be loaded. If memory usage is a concern, this could be lowered to 0xD00000 and change FX_MEMORY_32K_MIRRORS to 0xB00000, which should still function.

This pull request adds support for a homebrew mapper (more discussion here https://github.com/mrehkopf/sd2snes/issues/271) that allows SuperFX ROMs to go up to 11MB by setting 0x7FD7 (ROM Size) to 0x0E, and also being able to use more SRAM by respecting SRAMSize, the SuperFX can only address the first 2MB of the ROM still, and only 128KB of SRAM, however the SNES gets access to a full 9MB of extra ROM memory and up to 896KB of SRAM.

I also removed the redundant if (CalculatedSize > 0x400000) check since it just seems to be copypasted code.

Hopefully this change could be merged into the main version!

Test ROM:

[rom(2).zip](https://github.com/user-attachments/files/27281309/rom.2.zip)


OV2:
Sorry for the long delay, thank you for the PR!

The CalculatedSize checks were from the last extension to 8MB (2bb772322084db3fd905d3f9dd13139dd7423950)

PR #1053, opened 2026-05-20
win32: fix save-slot corruption from SelectSave hotkey on bank > 0

### LINKED ISSUE #1052: Bug with the Save State numbering when using hotkeys
Hey all.

If you save state via the file -> save game position menu, the save extensions are numbered correctly.

But if you save state using the shortcut keys, and switch from bank 0 to bank 1 (via hotkey), the resulting saves instead of being numbered from .010 - .019, they are numbered .020 - .029.

So for example, if I save in the fifth slot in bank 1, I get a save with the extension .025 instead of .015

Image
<br>

Image

shanytc:
The "Select save slot N" hotkey stored an absolute slot index (bank * SAVE_SLOTS_PER_BANK + i) into GUI.CurrentSaveSlot, but every other call site treats CurrentSaveSlot as the in-bank slot (0..9) and re-applies the bank multiplier when computing the filename. After pressing SelectSave on bank >= 1, CurrentSaveSlot was left >= 10, so the next save/load went to slot bank*10 + (bank*10 + i) and the status bar's "save slot 0NN" no longer matched the file actually written (e.g. slot 5 in bank 1 produced .025 instead of .015).

Store just `i` and clamp the value on config load so existing corrupted snes9x.conf entries self-heal.

Fixes #1052.

### DIFF
--- a/win32/wconfig.cpp
+++ b/win32/wconfig.cpp
@@ -277,6 +277,9 @@ const TCHAR*	WinParseCommandLineAndLoadConfigFile (TCHAR *line)
 
 	S9xLoadConfigFiles(parameters, count);
 
+	if (GUI.CurrentSaveSlot < 0 || GUI.CurrentSaveSlot > LAST_SAVE_SLOT_IN_BANK)
+		GUI.CurrentSaveSlot = 0;
+
 	ReleaseMutex(configMutex);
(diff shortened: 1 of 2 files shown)


shanytc:
Saves are now working as intended:

image

PR #1056, opened 2026-06-08
SuperFX: Fix RPIX Clearing DREG Too Early

### LINKED ISSUE #1055: SuperFX RPIX Clears DREG Too Early
SuperFX's RPIX instruction is supposed to write to DREG, according to all documentation I could find and the BSNES source. However, SNES9x clears DREG before this occurs. This causes the result to always be written to R0 and also causes TESTR14 to always fail.

The fix is easy: just move CLRFLAGS to be after TESTR14 in all three versions of the function.

Wyatt-James:
Fixes #1055

### DIFF
--- a/fxinst.cpp
+++ b/fxinst.cpp
@@ -751,7 +751,6 @@ static void fx_rpix_2bit (void)
 	uint8	v;
 
 	R15++;
-	CLRFLAGS;
 
 #ifdef CHECK_LIMITS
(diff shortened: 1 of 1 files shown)


bearoso:
There's a slew of problems with the GSU chip because it was originally just transcribed to C from zsnes. It needs to be replaced completely.

PR #1060, opened 2026-07-03
win32: recreate render window on display reset to fix Vulkan→D3D/DDraw switching

shanytc:
## Problem

On the Windows port, switching the output method from Vulkan back to Direct3D or DirectDraw freezes the game display (the last Vulkan frame stays on screen) while emulation and audio keep running; only restarting the application recovers. Reported and reproduced on NVIDIA hardware in shanytc/snes9x#152 (repro: load a game, switch Display Settings → Output Method to Vulkan, then back to Direct3D or DirectDraw).

This is documented Windows behavior rather than a driver teardown bug: once a **flip-model swapchain** has presented to an HWND, DWM permanently stops composing **blt-model presents** (windowed D3D9, DirectDraw, GDI) to that window — the HWND has to be destroyed and recreated. Current NVIDIA/AMD/Intel drivers route windowed Vulkan presentation through DXGI flip model, so after Vulkan presents once to the main window, subsequent windowed D3D9/DirectDraw presents succeed but are never composed. (That's also why the bug doesn't reproduce on machines where Vulkan never actually presents.)

## Fix

All display drivers now present into a dedicated borderless child window covering the client area, and `WinDisplayReset()` destroys and recreates it on every reset, so each renderer switch gets a fresh, untainted HWND — the same architecture other multi-API frontends use.

- **win32_display.cpp** — new `WinRecreateRenderWindow()` plus a minimal child wndproc: hit-test transparent (`WM_NCHITTEST` → `HTTRANSPARENT`, so mouse/superscope input still reaches the main window), forwards `WM_DROPFILES` to the main window, repaints via `WinRefreshDisplay()`. Also restores `WS_CLIPCHILDREN|WS_CLIPSIBLINGS` in the `ToggleFullScreen` style switches that previously dropped them.
- **wsnes9x.cpp** — the child tracks the client area on `WM_SIZE` before the driver's `ChangeRenderSize` runs.
- **CDirect3D.cpp/.h** — the device *focus* window stays the top-level `GUI.hWnd` (a D3D9 requirement for exclusive fullscreen); the *device* window is the child while windowed and swaps to the top-level window on the fullscreen `Reset`; the minimize check now tests `IsIconic(GUI.hWnd)` since a child window is never iconic.
- **CDirectDraw** needs no changes — the clipper receives the child through `Initialize()`, and the blt rects computed from `GUI.hWnd` are identical because the child exactly covers the client area.

A side benefit: OpenGL's once-per-window `SetPixelFormat` limitation now always hits a virgin window (see the existing workaround comment in `COpenGL::Initialize`). The existing OpenGL⇄Vulkan restart guards in `WinDisplayReset` are left untouched, though they may now be removable since every switch gets a new HWND.

## Testing

- Vulkan → Direct3D and Vulkan → DirectDraw while a game runs: video keeps updating (previously froze until app restart).
- Repeated back-and-forth switches between all four output methods, window resizing, borderless and exclusive fullscreen per renderer, ROM drag-and-drop onto the window, on-screen messages on each renderer.

### DIFF
--- a/win32/CDirect3D.cpp
+++ b/win32/CDirect3D.cpp
@@ -58,6 +58,7 @@ CDirect3D::CDirect3D()
 	shaderTimeStart = 0;
 	shaderTimeElapsed = 0;
 	frameCount = 0;
+	hWnd = NULL;
 	cgContext = NULL;
 	cgAvailable = false;
 	cgShader = NULL;
@@ -91,6 +92,8 @@ bool CDirect3D::Initialize(HWND hWnd)
 		return false;
 	}
 
+	this->hWnd = hWnd;
+
 	memset(&dPresentParams, 0, sizeof(dPresentParams));
 	dPresentParams.hDeviceWindow = hWnd;
     dPresentParams.Windowed = true;
@@ -100,7 +103,7 @@ bool CDirect3D::Initialize(HWND hWnd)
 
 	HRESULT hr = pD3D->CreateDevice(D3DADAPTER_DEFAULT,
                       D3DDEVTYPE_HAL,
-                      hWnd,
+                      GUI.hWnd,
                       D3DCREATE_MIXED_VERTEXPROCESSING,
 					  &dPresentParams,
                       &pDevice);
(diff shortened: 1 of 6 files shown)


shanytc:
@OV2 
seems like the build was stuck on zlib dependency, this ci job needs to re-run.

OV2:
Thanks. I've removed the OpenGL⇄Vulkan restart guards, this seems to fix those problems as well.

shanytc:
@OV2 awesome!

PR #1061, opened 2026-07-11
Qt niceties

YakoYakoYokuYoku:
This PR contains some improvements to the Qt GUI, namely a desktop entry and the posibility to pass ROMs via a positional argument.


bearoso:
Something you (and @shanytc) haven't noticed is that the Qt port is emulator-agnostic. The only part that should connect to Snes9x itself is in Snes9xController. It also works with other emulators using uncommitted controller classes.

shanytc:
> Something you (and @shanytc) haven't noticed is that the Qt port is emulator-agnostic. The only part that should connect to Snes9x itself is in Snes9xController. It also works with other emulators using uncommitted controller classes.

as I see it, if it's in snes9x project, there is no need to make it emu agnustic/frontend. why would you make something gerneric is a specific emulator repository?

bearoso:
> why would you make something generic is a specific emulator repository?

Separation of concerns. This makes it easier to eventually make gross changes to Snes9x's interface, which uses so many global symbols, and significantly needs it.

Remember I wrote the Qt GUI just for myself, and only added it in later without advertisement for those that might be interested. It's fine if you want to make changes to it. In this repository, though, I'm going to keep the core code going through the platform layer. I.e. for this pull request, don't include "snes9x.h" into the GUI for VERSION. Ideally, it would have a function to ask the compatibility layer for the core version.

bearoso:
I added the command line option. I don't want to add all the desktop files yet, because the Qt port isn't stable or complete, and I'd rather not encourage people to use it. Plus, instead of links, we'd want to just move those icons to the base data directory, since they're port-agnostic. I also just completely removed the GuiPrivate stuff from cmake, since we don't use it anymore.

PR #1065, opened 2026-07-18
SuperFX: cycle-accurate GSU timing; fix Winter Gold skier garble (#533)

shanytc:
## Summary

Fixes the long-standing *Winter Gold* garbled-skier bug (#533) by giving the GSU a real cycle-cost timing model and correcting a one-frame publish-visibility divergence at the game's race-start CPU<->GSU handshake.

## Root cause

Winter Gold runs a frame-quantized CPU<->GSU conversation: the CPU stages a pose message into a GSU queue each game-frame, the GSU renders and publishes the pose cel to `$70:EBC0`, and the CPU polls that byte once per frame (V~159) — nonzero commits the skier arrangement, zero parks the skier (blank frame).

Comparing the full pipeline frame-by-frame against a Mesen2 trace (staging cadence and V positions, publish frames, relay, OAM commits, tile uploads), everything matches except one value: at the race-start pose switch, snes9x's batch GSU model makes the publish CPU-visible **one frame earlier than hardware**. Hardware's poll still reads 0 (GSU busy with the course-load render) -> skier parked one frame, and the first tuck arrangement lands together with its tiles. snes9x's poll already sees the cel -> a **stale arrangement is committed over the new pose tiles** -> the garble. Verified directly: holding that byte at 0 for the single poll frame renders every frame clean.

## Changes

1. **Cycle-accurate GSU timing** (`fxinst.h`, `fxinst.cpp`, `fxemu.cpp`): `fx_run` now consumes a master-clock cycle budget per scanline (`Timings.H_Max`) instead of the legacy flat instruction count (magic 5823405/s). Per-op costs derived from Mesen2's GSU: cached fetch 1 (2 @10MHz), uncached fetch / RAM-ROM data access 5 (6), 16-byte cache-line fill 16x, amortized plot/rpix bitplane costs, multiplies per CFGR MS0. `snes9x_overclock_superfx` remains a true throughput knob (100 = hardware); `FXCYC=0` env selects the legacy budget for A/B.
2. **Publish-visibility shim** (`fxemu.cpp`, gated on ROM name `FX SKIING`): the 0->nonzero transition of `$70:EBC0` is held from CPU view for one frame (312 lines), matching measured hardware timing. The transition only occurs in race-start handshakes; steady gameplay never passes through 0, so the shim is inert elsewhere.

## Verification

- Reference savestate: previously garble at f153-155 / f159-161 / f165-174; now one blank frame at the switch (matching Mesen2 / hardware), then a clean animated skier on every frame; steady race clean; game cadence and switch frame unchanged.
- Confirmed in the win32 build.

## Regression notes

The cycle model changes effective GSU pacing for all SuperFX titles (cache-hot code faster than the legacy budget, memory-heavy code slower — both closer to hardware). Star Fox, Yoshi's Island, Stunt Race FX, Doom, Dirt Trax FX should get a play-test before merging into anything user-facing.

### DIFF
--- a/fxemu.cpp
+++ b/fxemu.cpp
@@ -33,6 +33,16 @@ void S9xResetSuperFX (void)
 	SuperFX.oneLineDone = FALSE;
 	SuperFX.vFlags = 0;
 	CPU.IRQExternal = FALSE;
+
+	// Cycle-accurate GSU timing (default). FXCYC=0 selects the legacy flat
+	// instruction budget for A/B comparison.
+	{
+		const char	*e = getenv("FXCYC");
+		g_gsuCycleMode = (e && *e == '0') ? 0 : 1;
+		g_gsuCycles = 0;
+		g_gsuCacheMask = 0;
+	}
+
 	FxReset(&SuperFX);
 }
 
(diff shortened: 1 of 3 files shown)


ReyVGM:
Finally, Star Fox's intro plays at the right speed!  Or at least 99.9% close to it.

vicde:
No bugs were found during the entire first world playthrough of Yoshi's Island.

shanytc:
@OV2 test it out when u get the chance. 
and yes, it also has a specific FX SKIING targeted fix for a specific intro <=> game frame issues, nothing to do with GSU.

ReyVGM:
> No bugs were found during the entire first world playthrough of Yoshi's Island.


YI never had any Super FX issues as far as I know.

OV2:
Still have not had any time to take a look, but I'll get to it eventually.

OV2:
I moved the global vars into the corresponding structure and aligned the game specific handling to the way it works for other games. Disabling is now via config setting or command line parameter.

PR #1066, opened 2026-07-18
Continue command for debugger

YakoYakoYokuYoku:
This adds a continue command for the Snes9x debugger. I had to change the binding of the palette dump command to follow the conventional continue binding for debuggers.


ReyVGM:
Wait, snes9x has a debugger? Where is it?

PR #1078, opened 2026-08-07
Win32: fix WaveOut sound-sync pacing locking ~20% below real time

shanytc:
## Problem

With the WaveOut driver and Sync Sound enabled, a pause/resume perturbation (e.g. opening and closing a menu, which drains the audio queue) can permanently drop the emulator to ~47-49fps with rhythmic audio gaps, until restart. Release builds only in practice — debug timing damps the oscillation — and XAudio2 is unaffected.

## Root cause

The sound-sync throttle paces the emulator on winmm `WOM_DONE` callbacks, but winmm delivers those from an internal pump that can lag the true play cursor by a full period. After the queue drains and refills, the loop can settle into a stable bad phase:

- each sync wait blocks on *callback delivery* rather than *actual playback* (~16ms instead of ~5ms per wait),
- the device silently starves between bursts while `bufferCount > 0` keeps `RecoverFromUnderrun()` blind,
- and the emulator locks to the callback phase at ~80% speed.

Diagnosed by instrumenting `ProcessSound()`: in the degraded state the emulator produced the correct 1066 words/frame but the device completed only 100 of 125 blocks/s, with the accounting never reading empty.

## Fix

- `GetAvailableBytes()` derives in-flight audio from `submittedBytes - waveOutGetPosition()` (wrap-safe), so pacing follows the device, not callback delivery. Because `waveOutGetPosition` round-trips into winmm and `ProcessSound` can run at scanline rate (e.g. an MSU-1 game min()-gates `S9xGetSampleCount()` and keeps the landing threshold satisfied, driving ~15k calls/s), the query is gated behind a cheap `bufferCount`-based bound that can only *under*-estimate free space — the cursor is read only when that bound says the queue may be full, or inside the sync wait itself.
- The sync wait rechecks free space after `ResetEvent` (closing a lost-wakeup race: a `WOM_DONE` landing between the check and the reset had its wakeup discarded) and caps each wait at 4ms; the 1s give-up is kept as total elapsed time.
- `bufferCount`/`submittedBytes` only advance when `waveOutWrite` succeeds — a failed write never gets a `WOM_DONE`, and counting it left a phantom entry that disabled underrun recovery forever.
- `SetupSound()` hard-resets the accounting (`bufferCount` was never initialized in the constructor nor reset across device re-opens).
- `RecoverFromUnderrun()` clears the stale partial block so pre-underrun audio no longer blips out after the silence cushion.

## Verification

Reproduced and verified on Windows 11 with an automated harness (Sound Settings OK + repeated menu open/close at varying cadence, FPS read from the on-screen counter): before the fix the degraded state latched reliably; after, 10/10 campaigns hold 60fps and per-second counters show the device fully fed (125 blocks/s at 48kHz/64ms) with zero dropped samples. The gated cursor query was additionally validated against a min()-gated secondary stream (thread stack sampling had shown the emulator thread saturated inside `winmmbase!waveOutGetPosition` without it).

### DIFF
--- a/win32/CWaveOut.cpp
+++ b/win32/CWaveOut.cpp
@@ -20,6 +20,10 @@ CWaveOut::CWaveOut(void)
 {
     hWaveOut = NULL;
     initDone = false;
+    bufferCount = 0;
+    submittedBytes = 0;
+    playedBytesAccum = 0;
+    lastPosBytes = 0;
 }
 
 CWaveOut::~CWaveOut(void)
@@ -59,6 +63,13 @@ bool CWaveOut::SetupSound()
 	int device_index = FindDeviceIndex(GUI.AudioDevice) - 1;
 
     waveOutOpen(&hWaveOut, device_index, &wfx, (DWORD_PTR)WaveCallback, (DWORD_PTR)this, CALLBACK_FUNCTION);
+
+    // Fresh device: position restarts at 0 and nothing is in flight, so
+    // reset the accounting instead of trusting waveOutReset's callbacks.
+    bufferCount = 0;
+    submittedBytes = 0;
+    playedBytesAccum = 0;
+    lastPosBytes = 0;
 
     UINT32 blockTime = GUI.SoundBufferSize / blockCount;
     singleBufferSamples = (Settings.SoundPlaybackRate * blockTime) / 1000;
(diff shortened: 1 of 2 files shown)


OV2:
I was so far unable to reproduce the issue on my system. Is this with default settings? Maybe also affected by sound driver?

shanytc:
@OV2 when you select waveout sound driver (click OK to confirm the sound settings dialog) - let the game run, click somewhere on the emu menu to show some menu (e.g Sound) and then release (click outside the menu to hide it) - it will cause stutter and lag so far as 20% below real time.

OV2:
I was unable to reproduce it on my system (even tried a few SS9X versions), so I guess it really depends on the system.

The free bytes loop seems a bit aggressive (missing a wakeup and falling one block behind shouldn't matter since the buffer is full), but I'll leave it as is for now.

shanytc:
> I was unable to reproduce it on my system (even tried a few SS9X versions), so I guess it really depends on the system.
> 
> The free bytes loop seems a bit aggressive (missing a wakeup and falling one block behind shouldn't matter since the buffer is full), but I'll leave it as is for now.

ss9x already has it implemented. hench why i created the PR. Thanks!

OV2:
I only used versions that do not have the fix to see if I was able to reproduce it, but was unable to get the low fps state.

Issue #318, opened 2018-06-01
Air Strike Patrol - interlace fixed, random crash after Marko + SGNG fix


kps501:
This one's been around for awhile. Start new game. Skip briefing 1. On radar area maps, pause emulator (single step frame). Minimize window. Restore window. If no garbage, single step frame and minimize - restore. You should now see some weird mini-map on top of the screen.

It switches mid-frame from 512 to interlace 512. memmove doesn't copy correctly but I think this one does.

gfx.cpp
				for (register int32 y = (int32) GFX.StartY - 2; y >= 0; y--)
					memmove (GFX.Screen + (y+1) * GFX.PPL,
							 GFX.Screen + (y+0) * GFX.RealPPL,
							 GFX.Pitch*sizeof(uint16));

bearoso:
@kps501, are you seeing lockups on Air Strike Patrol?

kps501:
No lockups. Just some strange visual junk when you pause the emulator, minimize and restore window. On the area radar briefing screen.

bearoso:
I'm definitely getting lockups after 85b0cf01919d254ec4a7c5fa919b474cbfa2c3f4.

kps501:
Same here. Just noticed it, that commit.

bearoso:
@OV2 any ideas? After reading more of anomie's and byuu's stuff, I removed the IRQTransition thing, because I think that's wrong. I also corrected a misinterpretation I had about vtimers, so I believe everything we're doing now is correct.

The behavior Marko expects is seemingly the reason ASP breaks. Marko wants CLI to not update the IRQ mask flag until after the next IRQ check, which is definitely the correct behavior. 

ASP does a CLI then SEI, which briefly unmasks the IRQ to trigger it, which seemingly works as far as my traces are concerned. The SEI then remasks it. I made a fix here where while the check for the IRQ flag at the end of SEI doesn't use SEI's result, it is still updated in time to be stored in the stack.

bearoso:
And I just noticed the SNES test program character test is broken.

OV2:
I'll try to take a look at ASP. I think the IRQTransition in the old variant was a way to simulate the cycle-delay between the counters being reached and the irq firing.

The character test was fixed by the openbus changes? I haven't really looked at how those work at all...

bearoso:
[...]

Issue #53, opened 2014-05-22
Snes9x Compatibility List


Oggom:
This has already been on the libretro wiki for quite some time, but I think it might be a good idea to post it here too:

A.S.P. Air Strike Patrol:
The shadow under the aircraft is missing.

Hayazashi Nidan Morita Shougi 2:
Matches won’t start.

Mecarobot Golf:
The ground “wobbles” during gameplay. Flickering graphics. 

Top Gear
Missing pixel garbage under the speed display. See https://github.com/snes9xgit/snes9x/issues/308

Fixed Issues:

Cho Aniki Bakuretsu Ranto-hen
~~Won't boot.~~ fixed as of https://github.com/snes9xgit/snes9x/commit/099bd60db0bec16b62a9351c32b24d3909ea55f2

Circuit USA
~~Broken graphics.~~ fixed as of https://github.com/snes9xgit/snes9x/commit/c22622806ded862eb5c9cbe9c040659ec80f1c69

Clayfighter 
~~The clownface turns blue when it should stay white..~~ fixed as of https://github.com/snes9xgit/snes9x/commit/1ca15b4a4f2cbbf3df64892eea41cdefc0a1ceb0

Daffy Duck: The Marvin Missions
~~Random screen flickering.~~ fixed as of https://github.com/snes9xgit/snes9x/pull/431/commits/2142282f69d7ee72d5ca851afe505536ff364ced

Dark Law
~~Garbage audio. Softlocks randomly.~~ fixed as of https://github.com/snes9xgit/snes9x/commit/fe6047213be802cb74cac724a1b83b7263c4dcca

Doom:
~~Colored dots appear during gameplay.~~ fixed as of https://github.com/snes9xgit/snes9x/commit/4345c20af2bcb15e3abfb3b3a796f2b3b6d991ba

Funaki Masakatsu Hybrid Wrestler – Tougi Denshou:
~~Corrupted graphics on the Pancrase logo screen.~~ fixed as of https://github.com/snes9xgit/snes9x/commit/9b595b5b197675f4b03a23794ec0d230224cf745

Joe and Mac 2
Often freezes when exiting a level on a dinosaur.
https://github.com/snes9xgit/snes9x/issues/660

JoJo no Kimyou na Bouken
~~Garbage characters show up during the intro~~ fixed as of https://github.com/snes9xgit/snes9x/commit/c96f6cd9138ad3c3f701b2e92c9864395e037677

[...]

PR #1079, opened 2026-08-09
PPU: render A.S.P. Air Strike Patrol's beam-raced mid-scanline effects

### LINKED ISSUE #524: A.S.P. Air Strike Patrol issue
In the A.S.P. Air Strike Patrol, a flashing horizontal line appears from time to time.

Trying the game with higan and a so-called "snesgt" the problem disappears.
https://i.postimg.cc/02dSZfsr/f1.jpg

Snes9x 1.59.2.

I seem to remember that in other versions it didn't happen.

### LINKED ISSUE #747: Air Strike Patrol graphic regression
Hello, I would like to report a problem that I found in the new version of Snes9x 1.61 that was released recently, both the standard version and the libretro version, and in the tests I ended up finding a graphical regression in the game Air Strike Patrol, in the message of "Good Luck", in which the rotation effect in this message is not working and so it is only locked and duplicated, and also the ship's shadow effect disappeared completely, before it appeared a little and blinking.  And so I think there may be problems in other games probably.

shanytc:
Fixes #524, fixes #747, and closes the A.S.P. Air Strike Patrol entry of #53.

A.S.P. (and its EU/JP release Desert Fighter) drives several effects from a per-line IRQ that races the beam, writing PPU registers mid-scanline and restoring them before HBlank, so the write timing itself bounds each effect horizontally:

- **aircraft shadow** — INIDISP dimmed 15→7 for ~10 dots on ~5 lines, alternate frames (the 30Hz blink is authentic)
- **GOOD LUCK message** — per-line BG3VOFS offsets: the rotation animation, and the weave that folds the tilemap's two field-half copies into one message
- **pause-map transparency** — W12SEL/WOBJSEL flipped to inverted over the map box span on alternate lines: the famous stripes
- **map grow-in** — MOSAIC enabled over the box span per line, with an animating block size

The existing `Timings.RenderPos = 128` entry latched every line before any of those writes landed, so each line rendered the restore values: shadow hidden (that was the hack's stated purpose), message frozen and duplicated, map opaque.

**Commit 1** restores SETINI interlace in all BG modes (partial revert of ff1925cc). The mid-frame instability that motivated that restriction (#318) came from this game's per-line SETINI writes, which the second commit's machinery handles; the game's mode-5 front-end genuinely bands interlace+pseudo-512 over scanline ranges via $2133 HDMA.

**Commit 2** drops the RenderPos entry and records mid-visible-line writes to INIDISP, the window-select bytes, MOSAIC and the BG scroll registers with their dot position. After the frame, each affected line's outer spans are re-rendered through the normal compositor with the pre/post-raster state, restricted via the clip-window segment lists and using the line's own HDMA-written window positions and backdrop color. Writes from HBlank onward (ordinary IRQ splits and HDMA) are excluded and stay on the per-line latch path, so the machinery is inert for games that don't race the beam — per the historical record, that is essentially this title alone.

Verified against bsnes-plus and Mesen references: shadow position/width and alternation, message animation with a stable HUD, stripes and mosaic bounded exactly to the map box. A ~90-ROM visual regression sweep reports no changes, and SMW / Star Fox / SF2 Turbo frame dumps are bit-identical with the patches applied.

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -117,7 +117,9 @@ void S9xGraphicsScreenResize (void)
 	IPPU.InterlaceOBJ = Memory.FillRAM[0x2133] & 2;
 	IPPU.PseudoHires = Memory.FillRAM[0x2133] & 8;
 
-	if (PPU.BGMode == 5 || PPU.BGMode == 6 || IPPU.PseudoHires)
+	// Interlaced frames render double-width too: the field-aware tile
+	// renderers live on the 512-wide path (see S9xSelectTileRenderers).
+	if (PPU.BGMode == 5 || PPU.BGMode == 6 || IPPU.PseudoHires || IPPU.Interlace)
 	{
 		IPPU.DoubleWidthPixels = TRUE;
 		IPPU.RenderedScreenWidth = SNES_WIDTH << 1;
@@ -456,7 +458,7 @@ void S9xUpdateScreen (void)
 			PPU.RecomputeClipWindows = FALSE;
 		}
 
-		if (!IPPU.DoubleWidthPixels && (PPU.BGMode == 5 || PPU.BGMode == 6 || IPPU.PseudoHires))
+		if (!IPPU.DoubleWidthPixels && (PPU.BGMode == 5 || PPU.BGMode == 6 || IPPU.PseudoHires || IPPU.Interlace))
 		{
 			// Have to back out of the regular speed hack
 			for (uint32 y = 0; y < GFX.StartY; y++)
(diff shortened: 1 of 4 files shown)


shanytc:
image

image

image

shanytc:
@bearoso this is a fun one

bearoso:
That's a novel trick to fix it post-rendering. It's not exactly ideal, which would be breaking scanlines up into spans that record volatile PPU state. 

An interesting thing about the master brightness change that bsnes/mesen don't take into account is that it takes about 10 dot cycles to take effect fully. So the shadow of the plane is lighter and has a very minor gradient to it on real hardware.

bearoso:
I'm gonna go ahead and merge this. I looked over it and it seems mostly fine. There's a bit of logical duplication, since it's segmenting the scanline separately both for max brightness changes and then again for the other PPU registers. We can clean that up later, though.

PR #1082, opened 2026-08-15
Add support for S-DD1 games converted to run without the chip

gufranco:
Street Fighter Alpha 2 and Star Ocean are the only S-DD1 cartridges, and both have conversions in circulation whose graphics were decompressed ahead of time so the chip is no longer needed. That is what lets them run from flash cartridges. Neither loads today.

The conversions grow the image and address it in two halves: the upper half of each bank sits where LoROM would put it, and the lower half sits one whole image further into the file. Banks $C0 and above are a window composed from the lower halves of two other banks, which is the part the existing map_lorom and map_hirom_offset helpers cannot express.

Detection has to happen before the HiROM and LoROM split. The Street Fighter Alpha 2 conversion scores as LoROM but Street Fighter Zero 2 scores as HiROM and would otherwise reach Map_ExtendedHiROMMap.

A conversion that keeps the retail header still declares the chip, so Settings.SDD1 alone covers those; no real S-DD1 cartridge exceeds Star Ocean's 48 Mbit, which makes 64 Mbit and above unambiguous. A conversion that corrects its header declares no coprocessor, so those are matched by cartridge name.

Verified against Street Fighter Alpha 2, Street Fighter Zero 2 and Star Ocean conversions at 96 Mbit, and against the retail cartridges of all three, which continue to use Map_SDD1LoROMMap.

### DIFF
--- a/memmap.cpp
+++ b/memmap.cpp
@@ -2248,7 +2248,18 @@ void CMemory::InitROM (void)
 	Map_Initialize();
 	CalculatedChecksum = 0;
 
-	if (HiROM)
+	const bool8	SDD1Decompressed = (CalculatedSize >= 0x800000) &&
+			(Settings.SDD1 ||
+			 strncmp(ROMName, "STREET FIGHTER ALPHA2", 21) == 0 ||
+			 strncmp(ROMName, "STREET FIGHTER ZERO2", 20) == 0 ||
+			 strncmp(ROMName, "Star Ocean", 10) == 0);
+
+	if (SDD1Decompressed)
+	{
+		Settings.SDD1 = FALSE;
+		Map_SDD1DecompressedMap();
+	}
+	else if (HiROM)
 	{
 		if (Settings.BS)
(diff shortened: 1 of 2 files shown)


niuus:
Can you share the patch or expanded Street Fighter Alpha/Zero 2 ROMs? I would like to bring this change to [Snes9xRX](https://github.com/niuus/Snes9xRX), but I don't have the ROMs to test. I googled a lot for these, but everywhere says they don't exist, except for Star Ocean 96Mb.

gufranco:
> Can you share the patch or expanded Street Fighter Alpha/Zero 2 ROMs? I would like to bring this change to [Snes9xRX](https://github.com/niuus/Snes9xRX), but I don't have the ROMs to test. I googled a lot for these, but everywhere says they don't exist, except for Star Ocean 96Mb.

https://github.com/shanytc/snes9x/commit/6fd4c71233b44f92879b94f6f655b12ec6980c4c#commitcomment-196601271

PR #1087, opened 2026-08-19
qt: make the SRAM autosave interval setting work

shanytc:
## Summary
Fixes the bug where the **Options → Files → "Save SRAM to disk every: N seconds"** setting in the Qt port neither persisted nor did anything. (GTK was checked and is wired correctly — this was Qt-only.)

The spinbox existed in `FoldersPanel.ui` but was orphaned at all three layers:

- `FoldersPanel.cpp` never connected the spinbox or loaded the stored value back into it, so the setting appeared to reset every time the dialog was reopened.
- `EmuConfig::setDefaults()` never initialized `sram_save_interval`, so an uninitialized value could be written to `SRAMSaveInterval` in the conf file.
- `Snes9xController::updateSettings()` never copied the value into `Settings.AutoSaveDelay`, which is what the core's autosave timer in `gfx.cpp` reads — so even a hand-edited conf value had no effect.

## Changes
- Connect `spinBox_sram_interval` to the config and refresh it when the panel is shown.
- Default `sram_save_interval` to 0 (save only on exit/reset), matching GTK and the core default.
- Apply the value to `Settings.AutoSaveDelay` in `updateSettings()`, clamping negatives to 0.
- Remove a leftover debug `printf` in `S9xAutoSaveSRAM()` that would print the .srm path on every periodic save now that the timer actually fires.

## Testing
- Built `snes9x-qt` successfully.
- Set an interval, closed/reopened the dialog — value persists; `SRAMSaveInterval` is written to the conf on exit; with a game that writes SRAM, the `.srm` mtime updates N seconds after an in-game save.


shanytc:
Opened against upstream by mistake, closing.

PR #1091, opened 2026-08-19
Name the expanded-cartridge map after its layout, and fix two defects in it

gufranco:
Follow-up to #1082. That PR was mine, and the naming I chose in it was wrong. I am sorry for the churn of coming back to the same code so soon.

I called the map `Map_SDD1DecompressedMap` and the flag `SDD1Decompressed`, and I gated it on `Settings.SDD1` plus three ROM titles. All of that says the layout belongs to the S-DD1, and it does not. It is a property of how a conversion re-addresses an image that outgrew the address space, and nothing in the mapping reads a chipset byte. The S-DD1 cartridges were simply the first two anyone had converted. I have since been expanding cartridges that trade a DSP-1 for a lookup table, and they land on exactly the same layout while declaring no coprocessor at all, which is what made the mistake obvious.

While fixing the name I found two defects in what I merged, one of which I introduced and one of which I missed. Both are described below with the measurements behind them.

## What changed

**Renamed to `Map_WindowedLoROMMap`.** The comment above it now describes the layout rather than the two cartridges it came from. No behaviour change in that commit.

**Recognised by shape rather than by title.** Plain LoROM starts mirroring above 4 MB, the extended layouts reach 8 MB, and no retail cartridge ever shipped larger than 6 MB, so an image past 8 MB in whole 64 KB banks and within the 12 MB the window addresses is one of these conversions whatever it dropped. The title list is gone rather than kept as a fallback: it was only needed because my size test was written as 8 MB *or more*, which an ExHiROM image can also be. Past 8 MB nothing else reaches, so the titles buy nothing.

Checked against 7,583 images on hand: the seven conversions are accepted and nothing else is.

**Kept these images out of the ExHiROM rearrangement.** This is the defect I introduced, and it is the serious one. Every image over 4 MB becomes an extended format unless its mapping and chipset bytes match an excluded pair, and the consequences run before `InitROM` chooses a map: `SMALLFIRST` rotates the ROM buffer in place, `BIGFIRST` reads the header from `0x407FB0`.

The two Star Ocean conversions escaped this only because they kept an S-DD1 chipset byte and `0x4532` is excluded. That protection is accidental, and it disappears the moment a conversion corrects its header to stop declaring a chip it does not have. The Street Fighter conversions never had it; they load today because `BIGFIRST` happens to land on a header mirror their builder placed at `0x407FC0`.

**Mapped save memory when a conversion declares it.** This one I missed in #1082. The map gave every bank but `$7E` and `$7F` to cartridge, which covers both LoROM save windows, and never called `map_LoROMSRAM`. Star Ocean declares 8 KB and both conversions inherited the declaration, so **Star Ocean has not been able to save since #1082 merged**. Guarded on the declaration, because `map_LoROMSRAM` claims `$70-$7D` whether or not there is any save memory and the Street Fighter conversions need those banks for cartridge.

I checked that this takes nothing away: in the Star Ocean conversion every one of those banks is 100% filler, because on hardware the cartridge decodes save RAM there and its builder left the space empty.

**Dropped every coprocessor the layout cannot reach, not just the S-DD1.** The map provides no coprocessor registers, so a chip the header still declares is one the cartridge cannot address. It matters most for a DSP, since `Map_LoROMMap` maps DSP registers over cartridge space. SuperFX and SA-1 are deliberately left alone: those run the game's own code rather than transforming its data, so expanding ahead of time cannot remove one.

## How it was tested

Built the libretro core from `master` and from this branch and ran the same images through both, comparing a signature of what each actually drew over 900 frames.

| Image | master | this branch |
|---|---|---|
| Star Ocean conversion, original header | 2,445 colours, save RAM **not mapped** | 2,445 colours, save RAM mapped |
| Star Ocean conversion, corrected header | **2 colours**, `SRAMSize` read as 90 | 2,445 colours, identical signature to the line above |
| Street Fighter Alpha 2 conversion | 323 colours | byte-identical |
| Street Fighter Zero 2 conversion | 355 colours | byte-identical |

A two-colour screen is a black one. The corrected-header image is rotated by the `SMALLFIRST` path on master before anything looks at it, and the header is then read out of the wrong place, which is where `SRAMSize` of 90 comes from.

The frame signature, lit-pixel count and pixel checksum of the corrected image on this branch are identical to the known-good original conversion, and both Street Fighter images are unchanged from master in every figure.

## Context, for anyone who wants to reproduce this

The conversions are not mine. The S-DD1 removal is [neviksti's Star Ocean no S-DD1/96Mbit hack](https://www.romhacking.net/hacks/614/), and the English one is built on DeJap's translation.

Two of my own repositories are what turned this up, in case they are useful to anyone testing:

- [street-fighter-alpha-2-nochip](https://github.com/gufranco/street-fighter-alpha-2-nochip) builds the Street Fighter Alpha 2 and Zero 2 conversions at 96 Mbit.
- [star-ocean-nochip-fix](https://github.com/gufranco/star-ocean-nochip-fix) corrects the Star Ocean conversions' headers, which is what exposed the `ExtendedFormat` problem: the images stop declaring a chip they no longer have, and lose the accidental exclusion that was keeping them off that path.

Neither distributes any image. Thank you for the review, and again, apologies for the naming in the first pass.

### DIFF
--- a/memmap.cpp
+++ b/memmap.cpp
@@ -2276,7 +2276,24 @@ void CMemory::InitROM (void)
 
 	if (WindowedLoROM)
 	{
+		// This layout provides no coprocessor registers, so one the header still
+		// declares is one the cartridge cannot reach. A conversion keeps the
+		// header of the cartridge it was made from unless somebody went back and
+		// corrected it, and plenty were not, so the byte routinely names the very
+		// part that expanding the data was what removed.
+		//
+		// SuperFX and SA-1 are left alone deliberately. Those run the game's own
+		// code rather than transforming its data, so no amount of expanding ahead
+		// of time removes one, and an image claiming either should not quietly
+		// pass for this layout.
+		Settings.DSP = 0;
+		Settings.C4 = FALSE;
 		Settings.SDD1 = FALSE;
+		Settings.SPC7110 = FALSE;
+		Settings.SPC7110RTC = FALSE;
+		Settings.OBC1 = FALSE;
+		Settings.SETA = 0;
+
 		Map_WindowedLoROMMap();
 	}
 	else if (HiROM)
--- a/memmap.cpp
+++ b/memmap.cpp
@@ -2248,16 +2248,16 @@ void CMemory::InitROM (void)
 	Map_Initialize();
 	CalculatedChecksum = 0;
 
-	const bool8	SDD1Decompressed = (CalculatedSize >= 0x800000) &&
+	const bool8	WindowedLoROM = (CalculatedSize >= 0x800000) &&
 			(Settings.SDD1 ||
 			 strncmp(ROMName, "STREET FIGHTER ALPHA2", 21) == 0 ||
 			 strncmp(ROMName, "STREET FIGHTER ZERO2", 20) == 0 ||
 			 strncmp(ROMName, "Star Ocean", 10) == 0);
 
-	if (SDD1Decompressed)
+	if (WindowedLoROM)
 	{
 		Settings.SDD1 = FALSE;
-		Map_SDD1DecompressedMap();
+		Map_WindowedLoROMMap();
 	}
 	else if (HiROM)
 	{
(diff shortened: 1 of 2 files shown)


gufranco:
@niuus, flagging this since you asked about #1082 with a view to bringing it into [Snes9xRX](https://github.com/niuus/Snes9xRX). Three things here are likely to matter to you, and one of them touches the exact ROM you said you were able to find.

**Star Ocean 96Mb could not save.** The map in #1082 handed every bank except `$7E` and `$7F` to cartridge, which covers both LoROM save windows, and never called `map_LoROMSRAM`. Star Ocean declares 8 KB and the conversion inherited that declaration, so the save window was never mapped. If you have already ported #1082, this is worth taking regardless of the rest.

**Detection no longer depends on a title list.** #1082 matched three ROM names, which would have meant maintaining the same list downstream and updating it for every new conversion. It is now decided by the image's shape, so there is nothing to keep in sync.

**A corrected header used to break the ROM.** Details are in the description above; the short version is that an image over 4 MB gets pulled into the extended-HiROM path unless its chipset byte happens to be excluded, and the Star Ocean conversions were only escaping that by still declaring an S-DD1.

On your original question, which I answered poorly at the time: I cannot share the Street Fighter images, but [street-fighter-alpha-2-nochip](https://github.com/gufranco/street-fighter-alpha-2-nochip) builds them from a copy of the game you own, so you would not be testing blind. [star-ocean-nochip-fix](https://github.com/gufranco/star-ocean-nochip-fix) does the same for the Star Ocean header correction, and pins every file in the chain by digest so you can confirm you ended up with the same bytes I did. Neither repository distributes anything.

Happy to answer anything that comes up while porting it.

PR #1092, opened 2026-08-20
Fix unzip includes for building w/o SYSTEM_ZIP

joshuamegnauth54:
The stream header and code files both had the wrong include for unzip.h. The rest of the files correctly include "unzip/unzip.h".


bearoso:
This isn't necessarily wrong. The build system is supposed to pull in root/unzip as an include directory. I'll go ahead and pull it anyway.

Issue #992, opened 2025-08-16
Error and segfault crash upon launching, possibly related to fontconfig


gnarlsnishi:
Using the AppImage release 1.63, on Arch Linux.

Upon launching Snes9x from the terminal, I get the following error message, and the app does not open.

`Fontconfig warning: using without calling FcInit()
Segmentation fault         (core dumped) ./Snes9x-1.63-x86_64.AppImage`

I also see the following line in my dmesg log.

`[ 7463.011177] AppRun[16267]: segfault at b8 ip 00007f308ca9f37b sp 00007ffcbb3d37e0 error 4 in libfontconfig.so.1.16.0[2737b,7f308ca7e000+31000] likely on CPU 0 (core 0, socket 0)`

I believe this is related to this issue with fontconfig: https://gitlab.freedesktop.org/fontconfig/fontconfig/-/issues/472

Thank you!

Tchou:
I confirm that it also happens on the latest ubuntu lts (26.04). Downgrading fontconfig fixes the issue.
I suggest adding an unaffected version of `libfontconfig.so.1` to the AppImage. I was able to fix it locally by:
- downloading the `.deb` of an older fontconfig : `libfontconfig1_2.15.0-2.3ubuntu1_amd64v3.deb`
- extracting `libfontconfig.so.1.12.1` from the package
- renaming it to `libfontconfig.so.1`
- extracting Snes9x-gtk.AppImage
- moving `libfontconfig.so.1` to `squashfs-root/usr/lib`
- repacking the AppImage

Alternatively, just doing `LD_PRELOAD=/path/to/old/libfontconfig.so.1 snes9x-gtk` also works.

jsnider3:
This still happens for me on Ubuntu 26.04  (GNOME 50, Wayland, kernel 7.0.0-30) with the official 1.63 AppImage. Claude was able to get it working. Here's it's diagnosis:

File tested — Snes9x-1.63-x86_64.AppImage, sha256 21a3d120dac7525a9effac3dbec82bd809dd91684bd46451c010480614427c39 (byte-identical to the GitHub release asset).

Symptom

Launches, then segfaults immediately on GUI init, preceded by:

Fontconfig warning: using without calling FcInit()

Kernel log:

AppRun[18872]: segfault at b8 ip 000070f453105953 sp 00007ffd69a10f40 \
  error 4 in libfontconfig.so.1.16.1[2d953,70f4530e0000+31000]

[...]

Issue #963, opened 2024-10-28
AppImage 1.63 doesn't start on Ubuntu 24.04


bglogic:
Trying to start Snes9x by running `Snes9x-1.63-x86_64.AppImage` on Ubuntu 24.04 gives the following error:

```
./Snes9x-1.63-x86_64.AppImage: error while loading shared libraries: libgiomm-2.4.so.1: cannot open shared object file: No such file or directory
```

I solved this by installing the `libglibmm-2.4-1t64` package:

```
sudo apt install libglibmm-2.4-1t64
```

A better fixed might be including `libgiomm-2.4.so.1` in the AppImage so it can work for everyone regardless of the linux distro.

the-jeab:
The same thing happened to me! I installed "libglibmm-2.4-1v5" on Debian ("libglibmm-2.4-1t64" does not seem to exist in its repositories) and it also solved the issue.

bearoso:
Duly noted. Sometimes it's hit or miss with dependencies. Include too many of the GUI ones and others will conflict, not to mention bloating the image. This alone would add at least 6 MB to it.

PR #1095, opened 2026-08-24
gtk: fix AppImage library bundling

jsnider3:
Fixes two library-bundling problems in `gtk/scripts/makeappimage.sh`. Related to #992 and #963.

## Problem 1 — the `libglib*` / `libgio*` globs also match the C++ bindings

`--exclude-library="libglib*"` matches `libglibmm-2.4`, and `--exclude-library="libgio*"` matches `libgiomm-2.4`. Both get dropped from the bundle even though the binary links against them, so the AppImage fails on any host without legacy glibmm installed — which is most current desktops:

```
AppRun: error while loading shared libraries: libgiomm-2.4.so.1:
cannot open shared object file: No such file or directory
```

This affects every build the script produces, on any distro. Anchoring the patterns to `libglib-2.0.so*` / `libgio-2.0.so*` fixes it while still excluding the C libraries as intended.

## Problem 2 — GTK bundled, fontconfig not

linuxdeploy applies the upstream AppImage excludelist, which drops `libfontconfig` and `libfreetype` by design (they read host font configuration). The script bundles `libgtk-3`, `libcairo` and `libpango*` anyway, so the font path is split across bundle and host.

Those libraries are tightly coupled, and a bundle built against one fontconfig can crash on a host with a newer one. The 1.63 release segfaults on Ubuntu 26.04:

```
Fontconfig warning: using without calling FcInit()
AppRun[18872]: segfault at b8 ip 000070f453105953 sp 00007ffd69a10f40 \
  error 4 in libfontconfig.so.1.16.1[2d953,70f4530e0000+31000]
```

Excluding the plain C GTK/pango/cairo/atk stack keeps the whole font path on the host side and internally consistent. The C++ bindings stay bundled, since those are what users are unlikely to have.

## Testing

Ubuntu 26.04, GTK 3.24.52, fontconfig 1.16.1, kernel 7.0.0-30.

| Build | Result |
|---|---|
| Released 1.63 AppImage | `libgiomm` missing; segfault in libfontconfig once that is installed |
| Released 1.63 bundle, GTK/pango/cairo removed by hand | runs |
| master @ 2971061, unpatched script | runs **only** because glibmm-2.4 was installed on the test host; the bundle omits it |
| master @ 2971061, **this patch** | runs |

Worth being explicit about row three: a fresh build on a current distro appears to work, because the bundled GTK then matches host fontconfig and the missing glibmm happens to be present. Neither condition holds for a release artifact on a user's machine, which is why the fix was also verified against the actual 1.63 bundle.

Bundle contents with the patch — `libgtk-3`, `libgdk-3`, `libcairo`, `libpango-1.0`, `libatk-1.0`, `libglib-2.0`, `libgio-2.0` absent; `libglibmm-2.4`, `libgiomm-2.4`, `libgtkmm-3.0`, `libgdkmm-3.0`, `libatkmm-1.6`, `libcairomm-1.0`, `libpangomm-1.4`, `libsigc-2.0` present. 45 libraries, 19 MB (unpatched: 52 libraries, 23 MB).

## Tradeoffs

- Problem 2's fix makes the AppImage require GTK 3 on the host. Present on essentially every desktop Linux, but it is a portability reduction — flagging it in case you'd rather go the other way. The Problem 1 fix has no such tradeoff and stands on its own if you prefer to take only that.
- The alternative for Problem 2 is force-bundling `libfontconfig`/`libfreetype` so the bundle is self-contained. I couldn't test that here: it needs a build on an older base to recreate the skew, and a bundled fontconfig reading a newer host's `/etc/fonts` has its own failure modes.
- Longer term, building the AppImage on the oldest supported distro would reduce skew generally, but that's a CI change rather than something this script can address.

### DIFF
--- a/gtk/scripts/makeappimage.sh
+++ b/gtk/scripts/makeappimage.sh
@@ -13,12 +13,12 @@ fi
 DESTDIR=AppDir ninja install
 ./linuxdeploy-x86_64.AppImage --appimage-extract-and-run --appdir=AppDir \
 	--exclude-library="libX*" \
-	--exclude-library="libglib*" \
+	--exclude-library="libglib-2.0.so*" \
 	--exclude-library="libgobject*" \
 	--exclude-library="libgdk_pixbuf*" \
 	--exclude-library="libwayland*" \
 	--exclude-library="libgmodule*" \
-	--exclude-library="libgio*" \
+	--exclude-library="libgio-2.0.so*" \
 	--exclude-library="libxcb*" \
 	--exclude-library="libxkbcommon*" \
 	--exclude-library="libdb*"
--- a/gtk/scripts/makeappimage.sh
+++ b/gtk/scripts/makeappimage.sh
@@ -21,7 +21,16 @@ DESTDIR=AppDir ninja install
 	--exclude-library="libgio-2.0.so*" \
 	--exclude-library="libxcb*" \
 	--exclude-library="libxkbcommon*" \
-	--exclude-library="libdb*"
+	--exclude-library="libdb*" \
+	--exclude-library="libgtk-3.so*" \
+	--exclude-library="libgdk-3.so*" \
+	--exclude-library="libcairo.so*" \
+	--exclude-library="libcairo-gobject.so*" \
+	--exclude-library="libpango-1.0.so*" \
+	--exclude-library="libpangocairo-1.0.so*" \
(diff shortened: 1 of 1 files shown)


bearoso:
I pushed the updated script I've been using. It doesn't do any additional exclusion.

jsnider3:
Thanks.

